Issue #438, opened 2026-08-16
Description narrows to a required field


ndelangen:
## Question

How does `description` narrow to a required `v.string()` once the backfill has run?

This is the second of the two PRs the field needs, and it exists as its own ticket because of a hard ordering constraint in `.github/workflows/deploy-main.yml`: `migrations:narrow-check` runs **before** `convex:deploy`, and it asserts against production that every migration id any *narrow* entry requires has already finished. Landing the narrow entry in the same PR as the widen would therefore fail the check on the very merge that was supposed to deploy the migration — the deploy deadlocks and nothing ships.

So this ticket may only merge after the widen release has deployed and `rulesets_description_v1` has completed in production.

The work:

- `convex/schema.ts`: `description: v.optional(v.string())` becomes `description: v.string()`, and the widen-phase comment on the field goes with it.
- `convex/migration-guards.json`: add `{ "id": "rulesets_description_narrow", "phase": "narrow", "requires": ["rulesets_description_v1"] }`.
- `src/shared/rulesets/validation.ts`: `description` stops being `.optional()` in `rulesetInputSchema`, unless the group-assignment ticket has already moved those callers off the general `update` mutation — check that first, because a required field there is what breaks them.
- `convex/rulesets.ts`: `parsed.data.description ?? ''` in `create` and the "absent means leave it alone" branch in `update` both stop being reachable and should go.

Verify before merging with `bun run migrations:narrow-check`, which is exactly what CI will run.

ndelangen:
**Now blocked by [Where the description is written](https://github.com/ndelangen/dunezone/issues/429) as well.** Making `description` required in `rulesetInputSchema` while no form collects one would leave the settings form unable to save at all — it would have to send a description it never gathered. So the authoring surfaces land first, then the narrow.

The other precondition this ticket named is **satisfied**: [#441](https://github.com/ndelangen/dunezone/pull/441) gives rulesets their own `setGroup`, so no caller reaches `update` without meaning to edit content, and `update` no longer accepts `group_id` at all. The checklist item about checking the group-assignment ticket first is done.

ndelangen:
**Unblocked** — both blockers are closed: the field exists (#439) and both forms now author it (#442).

[...]

Issue #429, opened 2026-08-16
Where the description is written


ndelangen:
## Question

Does the description get authored on the ruleset settings page, or inline on the detail page?

`RulesetSettings` in `src/app/routes/_app/rulesets/$rulesetSlug/edit.tsx` already holds the name and cover-image fields, so a `Textarea` there is the cheap answer and keeps one form owning one document.

Inline authoring on the detail page is the alternative: the description is the first thing in column one, and editing it where it is read is a shorter loop for a maintainer. The cost is a second write surface on a page whose every other write goes through the toolbar.

Whichever wins, the 50-character floor has to surface as a comprehensible message rather than a rejected submit.

ndelangen:
**Hazard inherited from the widen phase (#427).** `rulesetDescriptionSchema` has a 50-character floor and no allowance for `''`, so a form that echoes the ruleset's *current* description back on submit will fail validation for every row still holding the backfilled empty string. Codacy flagged this on #439 and suggested widening the schema to accept blanks; that was declined, because a floor that accepts blanks is not a floor.

The obligation therefore lands here: **send `description` only when the author actually wrote one.** An absent value means "not part of this write" and leaves the stored value untouched — that is the mechanism that keeps a rename from tripping the floor.

Also confirmed as part of this ticket's scope: the **create** form at `src/app/routes/_app/rulesets/create.tsx`, not just the settings form. Requiring 50 characters at creation is what breaks creation if there is nowhere to type them, so the field and the enforcement land together.

ndelangen:
## Resolution — all three fields on `/rulesets/x/edit`, floor enforced with no grace

Merged as `fd343740915` ([#442](https://github.com/ndelangen/dunezone/pull/442)); closed automatically by the PR.

**The answer.** The settings page owns the description alongside name and cover — not inline on the detail page. Name, cover and description are all ruleset content behind the same capability, the detail page routes every write through its toolbar, and an inline editor would have been the first exception in JSX that [The page re-cut](https://github.com/ndelangen/dunezone/issues/436) is about to rewrite.

**Scope that turned out to belong here:** the *create* form too. Requiring 50 characters at creation is what breaks creation if there is nowhere to type them, so the field and its enforcement had to land together.

[...]

Issue #427, opened 2026-08-16
Rulesets carry a description


ndelangen:
## Question

How does `description` reach prod as a required field without breaking the rows that predate it?

The sequence is widen → backfill → narrow, because `convex/migration-guards.json` is a real ledger with both phases and this is the reason the map ends in prod rather than in a merged PR.

- Widen: `description: v.optional(v.string())` on the `rulesets` table.
- Backfill: `` for every existing row. A generated placeholder was rejected — fake prose in the database is indistinguishable from authored text.
- Narrow: `description: v.string()`.
- Two ledger entries, `rulesets_description_v1` (widen) and `rulesets_description_narrow` (narrow), following the shape of the existing slug entries.

Validation lives in `src/shared/rulesets/validation.ts` beside `rulesetInputSchema`, and demands at least 50 characters on both create and update. Note that `rulesetInputSchema` is a `z.strictObject`, so adding a field there is not free — see the group-assignment ticket for the call site that breaks.

Settled: required, not nullable; `` only ever tolerated on read, for rows that predate the field.

ndelangen:
## Resolution — the widen half, implemented

`description` now exists on the `rulesets` table as `v.optional(v.string())`, with a backfill migration and the authoring floor in place. Implemented on `norbert/ruleset-factions-combo`, not yet merged.

**The sequencing finding that reshaped this ticket.** The charted scope bundled widen, backfill and narrow into one ticket. It cannot be one PR: `.github/workflows/deploy-main.yml` runs `migrations:narrow-check` *before* `convex:deploy`, and that check asserts against production that every id required by a narrow entry has already completed. A PR carrying both entries fails the check on the very merge meant to deploy the migration. The narrow is therefore its own release, now tracked as [Description narrows to a required field](https://github.com/ndelangen/dunezone/issues/438), which blocks [Shipped and verified in prod](https://github.com/ndelangen/dunezone/issues/437).

**What landed**

- `convex/schema.ts` — `description: v.optional(v.string())`, carrying a comment naming the phase it belongs to.
- `convex/migrations.ts` — `rulesets_description_v1`, idempotent: it patches `''` only where `description` is not already a string.
- `convex/migration-guards.json` — the widen entry, and only the widen entry.
[...]

Issue #426, opened 2026-08-16
Ruleset detail page: factions, description, and the re-cut layout


ndelangen:
## Destination

The reworked ruleset detail page, shipped to prod and verified there: rulesets carry a `description`, maintainers assign and unassign factions from the page itself, and the page is re-cut into a merged header plus three columns. Prod rather than a merged PR, because `description` is a schema change that needs two deploys.

## Notes

**Domain.** TanStack Router + Convex + Mantine. Read [`AGENTS.md`](../blob/main/AGENTS.md) before touching the app — the component taxonomy, the Picker contract, the Convex doorway and the shared-contract boundaries all bite here — and `convex/_generated/ai/guidelines.md` before touching Convex.

**This map carries execution.** Tickets end in merged code, and the last one ends in a verified prod deploy. Run `bun run worktree:setup` before any edit in a worktree; push to `real-origin`; run `bun run publisher:release:verify` before opening or updating a PR.

**Skills.** `/grilling` + `/domain-modeling` for decision tickets, `/prototype` for the header, `/tdd` where a seam suite is earned.

**Standing decisions from charting** — constraints for every ticket, not open questions:

- The many-to-many already exists. `ruleset_factions` is a join table, and `rulesets.addFaction` / `removeFaction` already enforce owner-or-active-group-member through `capabilities.edit`. No new permission model, no schema work on the relationship.
- The relationship is only ever edited from the ruleset side. The faction detail page keeps its read-only list of rulesets.
- One entry point for assignment: a toolbar button opening a popover that mounts the reused `FactionPicker`. Not two affordances.
- `description` is required in the schema, backfilled `''` for existing rows, and both create and update demand at least 50 characters.
- "Request membership" and the full membership badge leave this page — group membership is the group page's subject. A compact `StatusBadge` beside the maintaining group survives, showing *Member* or *Pending* and nothing otherwise.
- The fabricated "Version not specified" stat is deleted, not reworked.
- Stored faction data is assumed parseable. This page follows the catalogue's `parseStoredFactionForRead` and drops its own degradation branch.
- The factions column shows `FactionCard`. Its ruleset caption is suppressed by handing it an empty `rulesets` array, so no membrane change is needed for that.

## Decisions so far

[...]

PR #439, opened 2026-08-16
Add ruleset description field (widen phase)

ndelangen:
Adds `description` to rulesets as the **widen half** of `rulesets_description_v1`. The field is optional here on purpose; it narrows to a required `v.string()` in a second release.

Resolves the widen scope of #427. Part of the map in #426.

## Why two releases

`.github/workflows/deploy-main.yml` runs `migrations:narrow-check` *before* `convex:deploy`, and that check asserts against production that every migration id required by a `narrow` guard entry has already completed. A PR carrying both the widen and the narrow entry therefore fails the check on the very merge that was supposed to run the migration, and the deploy deadlocks. So this PR ships only the widen entry, and the narrow follows in #438 once `rulesets_description_v1` has completed in production.

## What changed

- **`convex/schema.ts`** — `description: v.optional(v.string())`, commented with the phase it belongs to.
- **`convex/migrations.ts`** — `rulesets_description_v1`, idempotent: patches `''` only where `description` is not already a string.
- **`convex/migration-guards.json`** — the widen entry, and only the widen entry.
- **`src/shared/rulesets/validation.ts`** — `rulesetDescriptionSchema`: `.trim().min(50)`, with no upper bound. `faqQuestionSchema` and `faqAnswerSchema` set the precedent that free prose carries no invented maximum. `description` joins `rulesetInputSchema` as optional for this phase only.
- **`convex/rulesets.ts`** — `create` and `update` accept `description` as an optional argument. An absent value means *not part of this write*, never *blank it*, so the ruleset toolbar's group assign and clear calls — which reuse `update` as a general patcher — cannot wipe a description. `create` writes `''` when none is supplied, matching what the backfill gives older rows.
- **`src/app/db/rulesets.ts`** — `Ruleset` gains `description?`, and both hooks parse the whole input instead of picking `name` out of it.
- **`convex/rulesets.description.test.ts`** — three contract assertions, each covering something types cannot express: a ruleset created without a description carries `''`; an update that omits it leaves the stored one intact; a description under the floor is rejected.

## Deliberately not in this PR

- **Wire validators are untouched.** `rulesetValidator` is `docValidator('rulesets')`, derived from the schema, so the field reaches the wire contracts without a hand edit.
- **`rulesetDescriptionSchema` is not exported.** It has no consumer outside its module yet and `knip` fails on unused exports; #429 exports it when the form needs it for UX feedback.
- **Enforcement is not live.** Both mutations still accept an absent description, so a ruleset can still be created without one. #429 owns both authoring surfaces — the create form as well as settings — because demanding 50 characters at creation is what breaks creation if there is nowhere to type them.

## Verification

`typecheck`, `lint`, `format:check`, `knip`, `migrations:static-check`, `check:app-layout`, `check:css-orphans`, `check:convex-skip`, `publisher:release:verify`, and the full suite at 469 tests — all green.

## Summary by CodeRabbit

* **New Features**
  * Added optional descriptions to rulesets.
  * Descriptions are trimmed and must contain at least 50 characters when provided.
  * Ruleset updates preserve existing descriptions when no new description is supplied.

* **Data Updates**
  * Existing rulesets are automatically backfilled with empty descriptions where needed.

* **Bug Fixes**
  * Improved validation consistency for ruleset creation and updates.

### DIFF
--- a/convex/migration-guards.json
+++ b/convex/migration-guards.json
@@ -11,6 +11,11 @@
       "requires": []
     },
     {
+      "id": "rulesets_description_v1",
+      "phase": "widen",
+      "requires": []
+    },
+    {
       "id": "faq_item_slug_v1",
       "phase": "widen",
       "requires": []
--- a/convex/migrations.ts
+++ b/convex/migrations.ts
@@ -32,6 +32,7 @@ type MigrationRef = FunctionReference<'mutation', 'internal'>;
 const MIGRATION_IDS: Record<string, MigrationRef> = {
   groups_slug_v1: internal.migrations.groups_slug_v1,
   rulesets_slug_v1: internal.migrations.rulesets_slug_v1,
+  rulesets_description_v1: internal.migrations.rulesets_description_v1,
   faq_item_slug_v1: internal.migrations.faq_item_slug_v1,
   faq_item_tags_v1: internal.migrations.faq_item_tags_v1,
   profiles_from_users_v1: internal.migrations.profiles_from_users_v1,
@@ -163,6 +164,22 @@ export const rulesets_slug_v1 = migrations.define({
     }
     const slug = await resolveUniqueRulesetSlug(ctx, row.name, row._id);
     return { slug };
+  },
+});
+
+/**
+ * Backfills `description` so the field can narrow to a required `v.string()`.
+ * Empty is the deliberate value: a generated placeholder would be indistinguishable from prose someone wrote, and the
+ * 50-character floor in `rulesetDescriptionSchema` applies to what is written, never to what is inherited.
+ */
+export const rulesets_description_v1 = migrations.define({
+  table: 'rulesets',
+  batchSize: 50,
+  migrateOne: async (_ctx, row) => {
+    if (typeof (row as { description?: unknown }).description === 'string') {
+      return;
+    }
+    return { description: '' };
   },
 });
 
--- a/convex/rulesets.description.test.ts
+++ b/convex/rulesets.description.test.ts
@@ -1,0 +1,65 @@
+/// <reference types="vite/client" />
+// @vitest-environment edge-runtime
+
+import aggregateTest from '@convex-dev/aggregate/test';
+import { convexTest } from 'convex-test';
+import { describe, expect, test } from 'vitest';
+
+import { api } from './_generated/api';
+import schema from './schema';
+
+const modules = import.meta.glob('./**/*.ts');
+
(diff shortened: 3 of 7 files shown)


ndelangen:
## Review responses

**CodeScene — `update.handler` complexity 11 → 12 (threshold 9): fixed in ce692eb.**

The gate is advisory here, but the complexity underneath it was real and not load-bearing: the handler assembled its partial patch through three `if (x !== undefined)` branches plus a hand-written type restating the patch shape. That is now `rulesetUpdatePatch`, which leaves the handler with parse, authorize and duplicate-check, and lets the patch shape be inferred rather than declared twice. Note the method was already above the threshold at 11 before this PR, so this takes it below where it started rather than merely back to baseline. Behaviour is unchanged.

**Codacy, `src/shared/rulesets/validation.ts` (HIGH) — "allow the empty string, since the description is meant to be optional": declining.**

The premise is inverted. `description` is a **required** field with a deliberate 50-character floor; that was the explicit product decision (#427, and the standing decisions on #426). Empty is not an optional-field sentinel here — it is the transitional backfill value for rows that predate the field, tolerated **on read only** until the field narrows to a required `v.string()` in #438.

Taking the suggested `refine((val) => val.length === 0 || val.length >= 50)` would make `''` a permanently legal *written* value, which is the design that was considered and rejected: a floor that accepts blanks is not a floor. The pattern being followed instead is the one already established in this repo for exactly this situation — `CanonicalFactionStoredSchema` is deliberately wider than `FactionInputSchema` so historical rows stay readable while new writes must satisfy authoring semantics.

On the "prevents users from updating existing rulesets" concern: no current caller submits a description at all. `RulesetSettings` and the ruleset toolbar's group assign/clear both send `input: { name }`, and an absent `description` means *not part of this write*, so they are unaffected. The hazard the comment is pointing at is real but forward-looking — a future form that echoes the current empty value back on submit would hit the floor — and it belongs to the form ticket (#429), which has been annotated with it. The fix there is that the form sends `description` only when the author wrote one, not that the schema learns to accept blanks.

**Codacy, `convex/rulesets.ts` (MEDIUM) — "a description cannot be cleared once set": working as intended.**

Accurate observation, but not a defect: a required field with a 50-character floor is precisely one you cannot blank. Adding a `null` to mean "reset to empty" would reintroduce the writable-empty state described above. If clearing a description ever becomes a product requirement, it needs a decision on the map before it needs a mutation argument.

**Sourcery / Copilot** — both declined to review on account quota limits, not on the diff.

Issue #436, opened 2026-08-16
The page re-cut


ndelangen:
## Question

What does the body of the page become?

Column one: the real description, then the "Resources" card moved down from the details column as the future rulebook's placeholder. Column two: the FAQ, unchanged. Column three: the linked factions as a single column of `FactionCard`s.

What comes out:

- The whole "Rules and variants" section — four `Card`s of invented rule headings.
- The "About this ruleset" `ProposedContent` block, replaced by the real description.
- The details column, whose contents have gone to the header or to column one.

Rendering the description: plain text in a `Textarea`-authored field, consistent with FAQ answers, which are plain text everywhere in this app. Decide how multi-paragraph text keeps its line breaks, and render nothing at all when the value is ``.

ndelangen:
**The header design is signed off** in [The merged PageHeader](https://github.com/ndelangen/dunezone/issues/434); fold it in from `norbert/prototype-merged-page-header` (`d5591a2`), rewritten properly rather than lifted — the variant code was written under prototype constraints.

Two things to carry:

1. **Verify the `StatusBadge` beside the group.** Every prototype screenshot was signed out, so *Member* / *Pending* has never actually rendered. It is the one merged element nobody has seen.
2. **The residual 1.1px of baseline scatter** on the meta line is 14px labels against 16px names and counts. The structural fix is a `size` prop on `Stats` so the counts can render at 14px — a kit decision worth making here rather than nudging pixels.

ndelangen:
## Resolution — the page as specified, mostly by deletion

Merged as `7a9340ef53b` ([#449](https://github.com/ndelangen/dunezone/pull/449)). Net 159 lines removed.

**Header:** breadcrumb; the cover as a 32px `Avatar` beside the name; then one line of owner, maintaining group with a `UsersRound` glyph, a membership badge when the viewer has a standing worth naming, and the three counts. Variant C from [The merged PageHeader](https://github.com/ndelangen/dunezone/issues/434), rewritten under production constraints rather than lifted.

**Body:** description then the rulebook placeholder in the primary column; the FAQ untouched in the secondary; the linked factions as a single column of `FactionCard`s in the rail.

**Deleted:** "Rules and variants" and its four invented rule headings; the "About this ruleset" placeholder; the entire details column including the fabricated `value: '—'` version stat; and the "Request membership" button, which took `useGroupMembershipWorkflow` and `canRequestMembership` out of the route with it.

[...]

PR #442, opened 2026-08-16
Author ruleset descriptions on the settings and create forms

### LINKED ISSUE #429: Where the description is written

ndelangen:
Adds the Description field to both ruleset forms, enforcing the 50-character floor client-side.

Resolves #429. Part of the map in #426.

## Decisions this encodes

- **All three fields live on `/rulesets/x/edit`**, not inline on the detail page. Name, cover and description are all ruleset content behind the same capability, and the detail page routes every write through its toolbar — an inline editor would be its first exception, in JSX that #436 is about to re-cut.
- **Authorization is unchanged.** The name stays owner-only via `capabilities.rename` (the field is already `disabled` for members); description and cover remain editable by an active member of the maintaining group, exactly as the cover is today.
- **The floor applies to every save, with no grace.** A ruleset still holding the backfilled empty string cannot be saved until someone writes a description. That is deliberate, not an oversight.

## What changed

- **`src/shared/rulesets/validation.ts`** — `rulesetDescriptionSchema` and `RULESET_DESCRIPTION_MIN_LENGTH` are exported now that the forms consume them (AGENTS.md sanctions client-side parsing for UX feedback). The comment on `rulesetInputSchema` claimed no caller could be forced to write 50 characters to rename a ruleset — that is now false, and it says so instead.
- **Both forms** — a `Textarea` whose description carries a live character count, an error only once something short has been typed, and a submit disabled until the value clears the floor. An untouched empty field is explained by the requirement line rather than shouted at.
- **`e2e/ruleset-lifecycle.spec.ts`** — fills the new field. Without it the Create button stays disabled and the spec fails, so this is a required change rather than added coverage.

## Not here

No new unit tests. The floor itself is already asserted at the seam in `rulesets.description.test.ts`, and the forms are route-local composition with no story surface; re-asserting the same rule through React would duplicate it (ADR-0002).

`publisher:release:verify` not run: nothing in this diff is a renderer-manifest ingredient, and CI's `publisher_release` job checks it regardless.

## Summary by CodeRabbit

* **New Features**
  * Rulesets now require a description of at least 50 characters when creating or editing.
  * Creation and editing forms display character progress and clear validation feedback.
  * Save and submit actions remain unavailable until the ruleset name and description meet validation requirements.
  * Descriptions are now included when creating or updating rulesets.

### DIFF
--- a/e2e/ruleset-lifecycle.spec.ts
+++ b/e2e/ruleset-lifecycle.spec.ts
@@ -8,6 +8,10 @@ test('owner can create and delete a ruleset in a two-user flow', async ({ page, 
   const expectedSlug = uniqueName.toLowerCase();
   await page.goto('/rulesets/create');
   await page.getByRole('textbox', { name: 'Name' }).fill(uniqueName);
+  /* Creation requires a description of at least 50 characters, with no exemption, so the button stays disabled without one. */
+  await page
+    .getByRole('textbox', { name: 'Description' })
+    .fill('A lifecycle ruleset proving that creation and deletion behave for its owner.');
   await page.getByRole('button', { name: /^create$/i }).click();
   await expect(page).toHaveURL(new RegExp(`/rulesets/${expectedSlug}$`));
 
--- a/src/app/routes/_app/rulesets/$rulesetSlug/edit.tsx
+++ b/src/app/routes/_app/rulesets/$rulesetSlug/edit.tsx
@@ -1,4 +1,5 @@
-import { Anchor, Center, Group, Image, Stack, Text, TextInput, Title } from '@mantine/core';
+import { Anchor, Center, Group, Image, Stack, Text, Textarea, TextInput, Title } from '@mantine/core';
+import { RULESET_DESCRIPTION_MIN_LENGTH, rulesetDescriptionSchema } from '@shared/rulesets/validation';
 import { createFileRoute, Link, useNavigate } from '@tanstack/react-router';
 import { FormError } from '@ui/block/FormError';
 import { SlugRenameNotice } from '@ui/content/SlugRenameNotice';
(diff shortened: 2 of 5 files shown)


ndelangen:
## Review responses

**MEDIUM, `create.tsx` — server errors pinned to the Name input: fixed in `14e4fcc`.** This was a real defect and the review caught it cleanly. `error={createRuleset.error?.message}` sat on the name field, so a rejected *description* would have rendered under *Name*. Errors now go to `FormError`, matching the settings form.

**LOW — nitpick on `event.target.value`: taken.** Now `currentTarget`, consistent with the settings form.

**LOW — extract the field into a shared component: taken in part.** The half with real risk is the duplicated *sentence*; the floor itself cannot drift, since both forms build it from `RULESET_DESCRIPTION_MIN_LENGTH` in the shared schema. So the copy moved to a kit Content support module (`rulesetDescriptionHint`), which is where AGENTS.md already puts copy that turns data into words.

Declining the component extraction. Per AGENTS.md a shared assembly across two routes would be a **Widget**, and widgets there are an explicit last resort — "every widget is a concession", and the top-level ones carry stories. That is a folder, a component and a story to own twelve lines of field markup whose only genuine drift hazard is now shared. The remaining duplication is a `useState`, a `safeParse`, and a `Textarea`.

**Test suggestions: declining, with one worth explaining.**

The React-level items (character counter, error-hidden-until-typed) re-assert through a rendered component what the schema already guarantees; ADR-0002 asks for a type guarantee or an existing path before adding a test, and the floor is asserted at the seam in `rulesets.description.test.ts`.

The interesting one is *"editing an existing ruleset with an empty description requires 50 characters before saving"* — that is the sharpest consequence of this change. It is left out for a structural reason rather than a preference: after this PR, **no UI can produce a ruleset with an empty description**, so the test would depend on seeded fixture state, and it stops being reachable at all once #438 narrows the field and the seed has to supply one. A test whose precondition is scheduled for removal is not coverage worth having.

`ruleset-lifecycle.spec.ts` was updated because it *had* to be — creation is impossible without the field, so the spec would have failed on a disabled button. That is the e2e leg this change actually needs.

PR #443, opened 2026-08-16
Add a ColumnsWithRailLayout and move the ruleset page onto it

### LINKED ISSUE #430: The three columns: TriptychLayout or the hand-rolled grid
## Question

Does the re-cut page keep its own grid, or move onto the kit `TriptychLayout`?

`RulesetDetail.module.css` hand-rolls `.detailGrid` — three named grid areas with two media-query breakpoints. `TriptychLayout` already exists in the kit with `Left` / `Center` / `Right` slots and is responsive by container query, which is what `AGENTS.md` requires of every Layout that is not `PageLayout`.

The case for moving: the columns are changing contents anyway, the kit component exists, and the hand-rolled grid is the only place on this page still laying out by viewport.

The case against: the current grid has a deliberate two-column intermediate arrangement at 62em that a generic triptych may not reproduce, and re-cutting the layout and its responsive behaviour at the same time makes a regression harder to attribute.

Decide before the page re-cut lands, because the answer decides whether the CSS module survives.

ndelangen:
The ruleset detail page's three columns become a kit Layout with named slots and container queries.

Resolves #430. Part of the map in #426.

## Why a new Layout, not a `TriptychLayout` variant

`TriptychLayout` centres its slots vertically around a middle panel holding artwork — the homepage passes `<AnimatedLeaderToken />` to its `Center` slot, which is why `.center` is `place-items: center` and why `.centerFill` exists. The ruleset page's middle column is the FAQ: text a reader works down, top-aligned, beside a narrow rail with a floor width.

Same silhouette, different intent. A `variant` prop would have meant one component owning two alignments, two sets of ratios and two collapse behaviours — two layouts sharing a name. `TriptychLayout` and the homepage are untouched by this PR.

## What it owns

- **Named slots** — `Primary`, `Secondary`, `Rail`, following the house habit of naming slots by proportion or role (`Wide`/`Narrow`, `Sidebar`/`Content`) so nothing domain-specific leaks into the kit.
- **Container queries, not media queries** — mandatory inside `src/app/ui/layout` and enforced by `containerQueries.test.ts`. The old grid responded to the window; this responds to the room it is given.
- **A two-stage collapse, kept deliberately.** Three columns → the rail rises beside the secondary column while the primary holds its width → everything stacks. That middle stage was already shipped behaviour, and it exists because collapsing a narrow rail straight into the reading flow buries it below the content it was meant to sit beside.
- **Stories** — four, under the existing Layout root: the three arrangements plus the intrinsic-sizing stress case the other layouts carry.

## What this PR does not do

Column *contents* are unchanged. Re-cutting them — description and rulebook placeholder, Q&A, factions — is #436, so this stays a pure move and the two changes can be reviewed apart.

`.detailGrid`, `.primaryColumn`, `.communityColumn` and `.detailsColumn` are deleted from the route's CSS module; the remaining `@media` blocks there style the page header, which is viewport-scoped chrome.

## Verification

`typecheck`, `lint`, `format:check`, `check:css-orphans` (52 stylesheets, no orphans), the full suite at 469 tests, and the new story file under the Storybook runner — 4 stories, all rendering.

`publisher:release:verify` not run: nothing here is a renderer-manifest ingredient, and CI's `publisher_release` job checks it regardless.

## Summary by CodeRabbit

* **UI Improvements**
  * Redesigned ruleset detail pages with clearer primary, secondary, and rail sections.
  * Improved responsive behavior across desktop, tablet, and mobile screen sizes.
  * Preserved existing ruleset content, filtering, navigation, and actions.
  * Improved readability and content organization through more consistent spacing and alignment.

* **Testing**
  * Added responsive layout previews covering stacked, intermediate, three-column, and content-sizing scenarios.

### DIFF
--- a/src/app/routes/_app/rulesets/$rulesetSlug/index.tsx
+++ b/src/app/routes/_app/rulesets/$rulesetSlug/index.tsx
@@ -26,6 +26,7 @@ import { StatusBadge } from '@ui/content/StatusBadge';
 import { TopicIcon } from '@ui/content/TopicIcon';
 import { AssignPopover } from '@ui/control/AssignPopover';
 import { IconAction } from '@ui/control/IconAction';
+import { ColumnsWithRailLayout } from '@ui/layout/ColumnsWithRailLayout';
 import { PageLayout } from '@ui/layout/PageLayout';
 import { FaqList } from '@ui/list/FaqList';
 import { Stats } from '@ui/list/Stats';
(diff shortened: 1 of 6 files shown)


ndelangen:
## Review responses

**CodeRabbit — slots hidden inside a fragment: correct, and fixed in `77889ff`.** This was a real bug and the best catch on this PR. Three of the four stories hoisted the slots into a shared `const slots = <>…</>`, and `Children.forEach` walks arrays but not fragments — so `ThreeColumns`, `RailBesidePrimary` and `Stacked` were rendering **three empty columns**. Worth naming why it slipped through: the Storybook runner smoke-renders, so all four stories reported passing while showing nothing. Green there meant "did not throw", not "shows content".

Fixed at the call sites rather than in the component. The stories now pass slots directly, which is what `TriptychLayout`, `AtlasLayout` and `AsymmetricSplitLayout` all expect; teaching this one layout to recurse into fragments would have made it the only permissive member of the four. A unit test now asserts each slot renders its children in order — and I checked it fails when the slots are fragment-wrapped, so it genuinely covers the bug rather than merely passing beside it. The constraint is also stated on the component, because the failure mode is silent.

That silent-empty behaviour is shared by all four layouts. Making it loud is a real question, but it is a kit-wide change and does not belong in a PR that adds one layout.

**Codacy (LOW) — abstract slot extraction into a `createSlotLayout` utility: declining here.** The duplication is real: four layouts now repeat the same `Children.forEach` walk. But that is a refactor of four existing kit components, and this PR adds one — folding it in would mean the layout that motivated the change and the change itself land together, with the homepage's layout in the blast radius. If it is worth doing it is worth doing on its own, where a regression has one obvious cause.

**Codacy (LOW) — early return on the last slot check: taken.** Free, and the loop reads uniformly now.

**Test suggestions:** all four listed behaviours are the four existing stories, and slot rendering is now unit-tested. Nothing further added.

**CodeScene approved. Sourcery and Copilot are both out of quota**, so neither is a signal.

Issue #435, opened 2026-08-16
Assign and unassign wired to addFaction and removeFaction


ndelangen:
## Question

How do the two dead mutations get a client surface?

`rulesets.addFaction` and `rulesets.removeFaction` already exist, already enforce owner-or-active-member through `requireRulesetMaintenance`, and are called by nothing. They need hooks in `src/app/db/rulesets.ts` and two affordances:

- **Assign**: one toolbar button, beside "Assign group", opening a popover that mounts the reused `FactionPicker`. The popover gates the mount, so the picker subscribes on mount and the inline shape stays unbuilt. Already-linked factions are excluded from the options.
- **Unassign**: a menu in each `FactionCard`'s `action` adornment, offering "Unassign from {ruleset}". No confirmation — one click through the same picker puts it back, unlike delete.

Both affordances appear only when `viewerAccess.capabilities.edit`, which is already owner-or-active-group-member. Errors surface where the page already surfaces mutation errors.

ndelangen:
**Two things inherited from [FactionPicker emits the faction id](https://github.com/ndelangen/dunezone/issues/432) ([#447](https://github.com/ndelangen/dunezone/pull/447)), both deliberately left for this ticket:**

1. **The generalisations the picker still needs.** Its copy is hardcoded to the editor (*"Load existing faction"*, *"Choose a faction first…"*, *"Replace this unsaved draft?"*) and its exclusion prop takes a single `currentPublicSlug` where linking needs every already-linked faction excluded. Both were left out of #447 because props with no consumer are unused code paths — this ticket is where the second caller lands, so it shapes them.

2. **A decision, not just work: does the authoring parse block linking?** The picker parses the chosen row with `FactionInputSchema`, which requires a name. `CanonicalFactionStoredSchema` — what storage actually permits — allows a blank one. So a historical faction with a blank name currently cannot be picked at all. That is correct for the editor (do not load an unparseable draft) and probably wrong for linking, where the faction exists and just wants attaching to a ruleset. Worth resolving with the linking UI in front of you rather than in the abstract.

The callback itself is done: `onPick` reports `{ id, slug, data }`, and `PickedFaction` is internal in the picker module until something imports it — this ticket likely will.

ndelangen:
## Answer

[...]

Issue #433, opened 2026-08-16
FactionCard gains an action adornment


ndelangen:
## Question

How does a menu sit in the card's top-right when the whole tile is a link?

`FactionCard` wraps everything in an `UnstyledButton` rendered as a router `Link`, so a menu button placed inside it would be a button inside an anchor.

Agreed shape: the card takes an `action` adornment, rendered as a sibling of the link and positioned by `FactionCard.module.css`. Alignment stays inside the kit component rather than the caller layering pixels over it, and `AGENTS.md` sanctions this explicitly — an action stays `ReactNode` in every category and never counts against a slot budget.

Constraints:

- The catalogue and the editor preview pass nothing and must be visually unchanged.
- The adornment must not extend the link's hit area, and must not be reachable by a click that navigates.
- Stories cover the card with and without the adornment, per the story conventions.

ndelangen:
## Resolution — a sibling adornment, top-left

Merged as `39d25310db7` ([#448](https://github.com/ndelangen/dunezone/pull/448)); closed by the PR.

`FactionCard` takes an optional `action`, rendered as a **sibling** of the link inside a positioned root — a button inside an anchor is neither valid HTML nor clickable without navigating, and the root keeps the link's own bounds so its hit area never grows. Alignment lives in `FactionCard.module.css`, so no caller layers pixels of its own.

**Top-left, against the ticket's top-right.** That corner is occupied by `.factionToken` (`top: 3%; right: 3%`, 20% width) — the faction's own identity token — with the `ComplexityGlyph` at the caption's right, the hero bottom-right, and the leaders and name across the bottom. Top-left is the one empty corner, which is also why a control there reads as chrome. The alternative was shifting an identity element on every catalogue card to make room for something one page shows.

**Existing callers are untouched:** the catalogue and the editor preview pass no `action` and render as before, same accessible name on the link, same grid behaviour.

**One story, and it was checked against the mistake it guards.** Its `play` asserts the button is visible and *not a descendant of the link*. Moving the adornment inside the `UnstyledButton` makes it fail with `expected true to be false`, so it covers the invariant rather than passing beside it — the lesson from the fragment bug in #443.

Issue #432, opened 2026-08-16
FactionPicker emits the faction id


ndelangen:
## Question

How does the picker hand back an id when its callback currently drops one?

`FactionPicker` resolves a selection to `structuredClone(parsed.data)` and calls `onLoaded` with it. `Faction` is `FactionInput` — the `data` payload alone, no `_id` and no slug, as the schema comment says outright. The row's id exists inside the component as `selectedRow.id` and never crosses the membrane.

So reuse means widening what the callback emits — id and slug alongside the parsed data — and updating the two existing callers in `factions/create.tsx` and `factions/$factionId/edit.tsx`.

Two things are hardcoded to the editor and have to become caller-supplied:

- The copy: "Load existing faction", "Choose a faction first…", "Replace this unsaved draft?".
- The exclusion: `currentPublicSlug` excludes exactly one faction, while the ruleset page needs every already-linked faction excluded.

The picker stays a Picker under the `AGENTS.md` contract: read-only, lazily mounted by its container, never mutating. The linking caller mounts it inside a popover the same way `FactionLoadPopover` does, so the inline-Picker shape stays unbuilt.

ndelangen:
## Resolution — the picker reports the row, not just the payload

Merged as `65a14728aed` ([#447](https://github.com/ndelangen/dunezone/pull/447)); closed by the PR. Two files, 18 lines.

`onLoaded` emitted `structuredClone(parsed.data)`, and a faction's id and public slug live on the **row**, never inside that payload — so the picker held `selectedRow.id` internally and discarded the answer at the membrane. It now emits `{ id, slug, data }` through `onPick`, named for what the control does rather than for what its first caller did with the result.

`FactionLoadPopover` keeps its editor-facing `onLoaded(faction)`, making it the single place identifiers are dropped: loading a draft needs the data, not the row. Neither route changed.

**Scoped to the callback on purpose.** The copy generalisation and the widened exclusion set were left for [Assign and unassign wired to addFaction and removeFaction](https://github.com/ndelangen/dunezone/issues/435), where the second caller lands — props with no consumer are the unused code paths AGENTS.md warns about, and `knip` made the same point unprompted by flagging `PickedFaction` as an exported type nobody imports. It is internal until something needs its name.

[...]

PR #446, opened 2026-08-16
Serve ruleset factions in the catalogue's shape

### LINKED ISSUE #431: Ruleset detail read model widens to catalogue-shaped factions
## Question

What does the ruleset detail query have to return for `FactionCard` to render?

Today `listPublicRulesetFactions` in `convex/lib/rulesetDetailPage.ts` returns a thin projection — `{ factionId, name, urlSlug, identity }` — which is enough for a `Spotlight` and nowhere near enough for `FactionCard`, whose membrane takes a `FactionCatalogueEntry`: fully parsed `data` (hero, leaders, complexity) plus a `rulesets` array.

The work:

- Return catalogue-shaped entries, parsed with `parseStoredFactionForRead` exactly as `convex/lib/factionCatalogue.ts` does.
- Delete `factionIdentityForClient` and the degradation branch it serves. The catalogue already lets an unparseable row throw, so this page tolerating one was the inconsistency.
- Hand each entry an empty `rulesets` array, which makes `factionRulesetLabel` return `null` and suppresses the card's ruleset caption with no change to `FactionCard`.
- Keep the read model's ownership intact: it owns slug resolution, the soft-delete gate, and viewer access, and the wire contract stays `rulesetDetailPageValidator`.

Watch the payload: this ships every linked faction's full `data` blob where it previously shipped four fields.

ndelangen:
The ruleset detail read model stops projecting linked factions into a bespoke four-field summary and sends whole catalogue-shaped factions instead.

Resolves #431. Part of the map in #426. Unblocks the `FactionCard` column in #436 and, with #432 and #433, the write path in #435.

## Why

`FactionCard`'s membrane takes a `FactionCatalogueEntry` — fully parsed `data` (hero, leaders, complexity) plus the faction's `rulesets`. The page was being handed `{ factionId, name, urlSlug, identity }`, which is enough for a `Spotlight` and nothing more. No amount of work in the route could close that gap; the read model had to widen.

## Reuse rather than a second shape

- **Wire contract** — `rulesetPublicBundleValidator.factions` now uses the existing `factionWithRulesetsValidator`, hoisted above its first use, rather than declaring a parallel shape. `rulesetFactionSummaryValidator` is deleted.
- **Client boundary** — `factionCatalogueRowsToEntries` was already exported from `@db/factions`, so `@db/rulesets` reuses it instead of keeping a second normaliser.
- **Parsing** — `parseStoredFactionForRead`, exactly as `convex/lib/factionCatalogue.ts` does it. The old degradation branch that returned a null identity for unparseable data is gone: the catalogue already lets such a row throw, so this page tolerating one was the inconsistency, not the safety net.

## `rulesets` comes back empty, deliberately

These factions are read *inside* one ruleset, so captioning each card with a ruleset name would repeat the page's own subject back at the reader. An empty array is also the mechanism: `factionRulesetLabel` returns `null` when the list is empty, which suppresses the card's caption with no change to `FactionCard` itself.

## Fallout, all of it deletion

- `BackgroundClientSchema` and the `FactionData` alias had one consumer each — the normaliser this PR deletes — so both are gone. `knip` named them.
- `Ruleset` in `@db/rulesets` still declared `description` optional, which stopped being true when the field narrowed in #445. It is now `RulesetInput`, derived from `rulesetInputSchema`, so the type cannot disagree with the schema that validates it. That mirrors how `Faction` derives from `FactionInput`.
- One assertion in `factions.authoringRoundTrip.test.ts` moved from `identity.background` to `data.background` — the same round-trip claim through the new shape.

Net 93 lines removed against 35 added.

## Not here

The column still renders `Spotlight`, adapted to the new field names. Swapping in `FactionCard` belongs to #436, so this stays a data change.

## Verification

`typecheck`, `lint`, `format:check`, `knip`, `check:convex-skip`, and the suite at 469 tests across 76 files — all green.

`publisher:release:verify` not run: nothing here is a renderer-manifest ingredient, and CI's `publisher_release` job checks it regardless.

## Summary by CodeRabbit

* **Improvements**
  * Faction information in ruleset pages now uses complete catalogue details, including names, logos, backgrounds, and links.
  * Ruleset data is now handled consistently across creation, editing, and display.
* **Validation**
  * Ruleset descriptions are now required when creating or updating a ruleset.
  * Invalid faction data linked to a ruleset is rejected instead of being displayed incompletely.
* **Bug Fixes**
  * Updated ruleset and faction data handling to maintain accurate information during loading and editing.

### DIFF
--- a/convex/factions.authoringRoundTrip.test.ts
+++ b/convex/factions.authoringRoundTrip.test.ts
@@ -296,10 +296,11 @@ describe('faction authoring full-field round trip', () => {
         },
       ],
     });
+    /* The ruleset page now carries whole catalogue-shaped factions, so the round trip is checked through `data`. */
     await expect(t.query(api.rulesets.getBySlug, { slug: 'canonical-transition-proof' })).resolves.toMatchObject({
       factions: [
         {
-          identity: {
+          data: {
             background: {
               invert: false,
             },
--- a/convex/lib/collaborativeAccessValidators.ts
+++ b/convex/lib/collaborativeAccessValidators.ts
@@ -2,7 +2,7 @@ import { v } from 'convex/values';
 import type { Infer } from 'convex/values';
 
 import schema from '../schema';
-import { factionBackgroundValidator, factionDataValidator } from './factionData';
+import { factionDataValidator } from './factionData';
 
 /**
  * Document validators derive from their authority, `convex/schema.ts` (ADR-0002);
@@ -118,16 +118,14 @@ export const factionDetailPageValidator = v.object({
   rulesets: v.array(rulesetSummaryValidator),
 });
 
-const rulesetFactionSummaryValidator = v.object({
-  factionId: v.id('factions'),
-  name: v.string(),
-  urlSlug: v.string(),
-  identity: v.union(v.object({ logo: v.string(), background: factionBackgroundValidator }), v.null()),
+/** A faction with the rulesets it belongs to: what both the catalogue and a ruleset's own page put on the wire. */
+export const factionWithRulesetsValidator = factionValidator.extend({
+  rulesets: v.array(rulesetSummaryValidator),
 });
 
 export const rulesetPublicBundleValidator = v.object({
   ruleset: rulesetValidator,
-  factions: v.array(rulesetFactionSummaryValidator),
+  factions: v.array(factionWithRulesetsValidator),
   viewerAccess: rulesetViewerAccessValidator,
 });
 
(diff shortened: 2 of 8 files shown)


ndelangen:
`publisher_release` failure fixed in `6f02e163` — the regenerated manifest is committed.

**My error, and the rule I had wrong.** I skipped `publisher:release:verify` locally on the grounds that nothing in this diff appears in `RENDERER_RUNTIME_CLOSURE_PATHS`. That list is not the test. The manifest's `code` component is:

```ts
const codeFiles = [
  ...filesBelow(publisherDirectory).filter(…),   // built capture-bundle bytes
  ...RENDERER_RUNTIME_CLOSURE_PATHS.map(…),
];
```

The first line is the built capture bundle, and the capture page imports `src/shared/factions/schema.ts` transitively — so deleting two unused exports from it changed the bundle bytes and moved the digest. A fixed path list can never answer this; the import graph does.

**It is identity, not staleness.** The regenerated hashes match what CI computed byte for byte — `digest: cd27d20d…`, `code: 3b128a16…` — which is exactly the cross-machine reproducibility the manifest is designed for, and the reason this is a local git-diff check rather than a CI-only artifact.

Nothing about the change itself is affected: the same two orphaned exports go, for the same reason.

Issue #569, opened 2026-08-20
Prod verification: every asset type live, published, and walked


ndelangen:
## Question

(task) The destination check, in production: every asset type creatable, editable, browsable, published, and actually looked at.

Create and publish one of each: a treachery card, the four token shapes (at least one with a custom back and one with a referenced back), a deck with its Cardback, a bundle with members. Verify the published bytes at their public URLs, not just job success: progressive JPEG (`FFC2` start-of-frame), correct dimensions per `PUBLICATION_TARGETS`, and the back face under its face-qualified id. Walk the journeys: the `/assets` landing with its category piles, each type browse page, detail pages including a deck's "used by these rulesets" and a bundle's members, edit round-trips, and ruleset slots (`techToken` single, `customTokens` many).

Two calibration points carried from the fog:

- JPEG quality: q88 was measured on fixture card art. If production art argues with it, the quality is a one-line change per type in `PUBLICATION_TARGETS`.
- «The publisher waits for three fonts and renders with seven» (#563) means a first card or token publish can capture before its faces load and publish fallback-rendered bytes. Look at the published images for the right faces before calling them verified. If this bites, #563 gates the result and is fixed on its own ticket, not this one.

The lesson this ticket exists to apply: on 2026-08-20 the first live pass of the day found two defects that every unit, story, and capture gate had passed, one of them invisible on the most likely test content. A green pipeline is evidence the contracts hold, never that anyone has seen the thing. This is HITL: creating production content and judging rendered output needs Norbert in the loop.

Resolution records what was created, the published URLs checked, and any quality or font finding.

ndelangen:
**Resolved. Every asset type is live in production, published, byte-verified, and walked; two product decisions came out of the walk.**

**Created** (all by Norbert, owner "Central"): treachery cards Trishula! and Supplies!; decks Supplies! cache and Dreamrules Treachery Deck; tokens Spice (disc), Water Stealing (tech), CHOAM Share (plate), Kwizatz Haderach and Immunity (enhance); bundle Dreamrules Techtokens; ruleset Dreamrules with its slots filled.

[...]

Issue #566, opened 2026-08-20
Deleting an account leaves its Assets behind


ndelangen:
## The gap

Deleting an account does not touch that account's Assets. They stay live, owned by a deleted user, and keep appearing under `/assets`.

`DIRECT_OWNERSHIP_KINDS` in `convex/lib/directOwnership.ts` names three kinds: `group`, `faction`, `ruleset`. `assets` is not among them, so `accountDeletion`'s snapshot walk never sees an Asset, and no `account_deletion_items` row is ever written for one.

## Why nothing caught it

**Neither branch was wrong on its own.** Account deletion was built on `main` while Assets were built on `norbert/assets`, and each was complete against the tree it could see. The gap came into existence at the merge, `13e6c59de5a`, and no test on either side could have failed for it.

Verified against the merged tree:

- `convex/lib/directOwnership.ts` lists three kinds, no `asset`.
- `convex/schema.ts:240` types `account_deletion_items.entity_id` as `v.union(v.id('groups'), v.id('factions'), v.id('rulesets'))`, exactly those three.
- `convex/accountDeletion.ts` drives its whole walk off `DIRECT_OWNERSHIP_KINDS`: the summary at line 156, the first snapshot kind at 237, and the kind advance at 321 and 330.

## What the fix costs, for whoever picks it up

The registry entry itself is nearly free. `assets` already has `owner_id`, `is_deleted`, and a `by_owner_deleted` index (`convex/schema.ts:83`), so it fits the existing shape with no schema change and no new index:

```ts
{ kind: 'asset', table: 'assets', ownerField: 'owner_id', deletedField: 'is_deleted', ownerDeletedIndex: 'by_owner_deleted' }
```

Three things around it are not free.

**`directOwnershipKindValidator` is hand-written as three explicit `v.literal(...)` arms** rather than derived from the array beside it. A fourth kind needs a fourth arm, and nothing fails if you forget. Worth deriving while adding to it, since this is the same second-source-of-truth shape that has already produced three defects on this branch today.

**`entity_id`'s union needs `v.id('assets')`**, which is a schema widening on a table that already holds rows, so it follows the widen-then-narrow discipline in `convex/migration-guards.json`.

[...]

Issue #563, opened 2026-08-20
The publisher waits for three fonts and renders with seven


ndelangen:
## Question

Nothing to decide about *whether*. `assertRequiredPublisherFonts` is the capture page's guarantee that a face has loaded before the driver is allowed to take bytes, and it currently covers three of the seven faces the project ships.

`REQUIRED_PUBLISHER_FONTS` in `src/shared/asset-publishing/publisher-fonts.ts`:

- Caladea, four faces
- `C_Copperplate_Gothic` 400 normal
- `C_Advokat_Modern` 400 italic
- `C_Trebuchet` 400 italic

Not asserted at all: **`C_Busorama`**, **`C_Desdemona`**, **`C_Candara`**, **`C_Copperplate_Gothic_Heavy`**. `C_Advokat_Modern` and `C_Trebuchet` are asserted only in italic, so their normal faces are uncovered too.

### Why it matters

All seven are declared in `src/app/styles/fonts.css`, and `publisher-entry.tsx` imports that file, so every face reaches the capture bundle. The gap is not availability, it is **readiness**. The capture page waits on `document.fonts.ready` and then on the required list; a face outside that list can still be loading when the marker flips to `ready`, and the driver's licence to screenshot is that marker. The failure mode is a published image rendered in a fallback face, which looks fine to every automated check and wrong to a human.

### This is not hypothetical, and it is not new

Existing `src/game` renderers already use all seven:

```
10  C_Copperplate_Gothic
 4  C_Desdemona
 4  C_Busorama
 3  C_Trebuchet
 2  C_Candara
 2  C_Advokat_Modern
 1  C_Copperplate_Gothic_Heavy
```

So the faction sheet and the treachery card carry this exposure **today**, in production. «[Build the rectangle token editor](https://github.com/ndelangen/dunezone/issues/523)» found it rather than caused it, and deliberately did not widen the list, because putting every existing capture behind four more font loads belongs in a ticket that can weigh that risk rather than one that publishes nothing.

### What makes it slightly more than a list edit

`assertRequiredPublisherFonts` **throws** when a required face fails to load exactly, and an encode failure fails the job. So adding faces makes every capture, including the faction sheet's, depend on those four loading. That is the right trade, but it wants the Chromium capture regression run before and after rather than a blind list edit, and it wants a decision about the weights and styles actually used rather than adding all seven at 400 normal and calling it covered.

### Done when

[...]

Issue #490, opened 2026-08-18
Wayfinder map: Assets — browse, edit, publish


ndelangen:
## Destination

Treachery cards, decks, and tech tokens in all four shapes (round, gear, square and rectangle) are live in production: creatable, editable, browsable under `/assets`, and published through the existing publication pipeline — cards and tokens as progressive JPGs, decks as their Cardback — deployed and verified in prod. The `/assets` landing page shows all Asset categories, with unbuilt ones presented as planned placeholders. The asset data model is locked so remaining categories (boards) and variations (other card templates) land later without rework.

## Notes

- Domain vocabulary lives in `CONTEXT.md` — this effort added **Asset type**, **Asset category**, **Cardback**, **Ruleset deck slot**. Consult it before every session.
- Grilling tickets: use `/grilling` + `/domain-modeling`. Prototype tickets: `/prototype` (explicit sign-off per sub-element; alignment via the layout system). Research tickets: `/research`.
- **Management actions sit in both toolbars**: delete and group assign/remove appear on the detail page *and* the edit page of every entity, each gated on the viewer's real capabilities (Norbert, 2026-08-20). App-wide as a pattern; this map applies it to what it builds rather than retrofitting the faction, ruleset and profile pages.
- **Destination redrawn 2026-08-20.** Norbert: *"I'd love to ship with all 4 tokens"*. The rectangle token editor had been ruled out of scope on the grounds that the destination named other token shapes a later variation; that clause is gone, and «[Build the rectangle token editor](https://github.com/ndelangen/dunezone/issues/523)» graduated into the map. What made it affordable is that «[Rectangle token editor](https://github.com/ndelangen/dunezone/issues/508)» left 656 working lines on its prototype branch, so the build productionizes rather than invents.
[...]

PR #570, opened 2026-08-20
Assets: browse, edit, publish (wayfinder #490)

ndelangen:
Everything «[Wayfinder map: Assets — browse, edit, publish](https://github.com/ndelangen/dunezone/issues/490)» built: 43 closed tickets, 67 commits.

## What ships

- The `assets`, `asset_relations`, and `ruleset_asset_slots` tables, with `assets.data` as Zod-validated `v.any()` following the faction pattern.
- Fourteen asset types in one registry: treachery cards, decks, four token shapes, bundles, and planned placeholders.
- The `/assets` landing (category piles), per-type browse pages, addressable detail pages for every type, and per-type create/edit routes with five editors (card, deck, round/gear/square token, rectangle token, bundle).
- Publication generalised over asset type: capture rides the snapshot envelope, publish-time progressive JPEG encode through the Images binding, per-type route/key/content-type table. Decks publish their Cardback; tokens publish both faces, the back under a face-qualified id.
- Ruleset asset slots: `treachery`/`spice`/`custom` deck slots plus `techToken`/`customTokens` bundle slots, managed on the ruleset edit page, read back on deck and bundle detail pages.
- Assets join collaborative access as their own subject kind. Owner-only delete and group reassignment on both toolbars.

## What the deploy does on merge

`deploy-main` runs the two `assets_about` widen migrations, then the renderer activation step flips six new types to revision 1 (`card-treachery`, `deck`, `token-disc`, `token-tech`, `token-plate`, `token-enhance`). Production has zero asset rows, so activation backfills nothing; it opens publication for the new types. The `dev_rebuild` data stage fires because this branch's diff touches `convex/schema.ts`, `convex/migrations.ts` and `convex/migration-guards.json`, which is the conditional path, not the default; a ship that leaves those files alone pushes only code. The rebuild wipes and re-seeds the dev deployment, which also clears the schema drift three branches collided over on 2026-08-20.

The renderer manifest digest moved with the token-type rename and is committed in that same commit (`eb5622d9…`); main touched nothing under `workers/publisher`, so the capture closure is this branch's alone.

## What this does not close

- «[Prod verification: every asset type live, published, and walked](https://github.com/ndelangen/dunezone/issues/569)» runs after this merges; the map stays open until it passes.
- «[The publisher waits for three fonts and renders with seven](https://github.com/ndelangen/dunezone/issues/563)» can make a first card or token publish capture with fallback faces. Known, filed, gates the verification rather than this merge.
- «[Deleting an account leaves its Assets behind](https://github.com/ndelangen/dunezone/issues/566)»: `DIRECT_OWNERSHIP_KINDS` does not cover assets. Found in the merge of main, filed separately.


ndelangen:
This PR becomes the integration reference: the same tree now lands as a reviewable stack — #580 (model, backend, faces) ← #581 (the publisher worker) ← #582 (the kit) ← #583 (pages and editors). Each slice is under the bot-review file limit and prod-safe to merge in order; the stack's final tree reconstructs this branch exactly, plus four picker exports made private when the knip ignore hiding them retired. This PR stays open for the whole-picture diff until the stack lands, then closes in the stack's favour.

PR #581, opened 2026-08-20
Assets 2/4: the publisher learns to make images

ndelangen:
Slice 2 of 4 of the assets stack, on top of #580.

The publisher worker: capture generalised over asset type, publish-time progressive JPEG through the Images binding, per-type storage and delivery from the targets table, face-qualified back publications, and the capture-contract regression driving faction, card, deck and token shapes.

**Review here**: the byte contract (PNG IHDR, JPEG SOF2, dimensions, size cap), the per-job error boundary, R2 pathing and cache tokens, the regression's coverage.

The renderer manifest digest is unchanged from the integration branch, which proves the slice boundary: the `code` hash covers the capture bundle's import graph, carried whole by slices 1+2; the app slices never touch it. Deploying alone is safe: activation flips six renderer types over zero rows; jobs only exist when saves create them, which needs slice 4.


ndelangen:
Closed until its base merges, to keep the review apps' attention (and rate budget) on one slice at a time. The branch stays pushed and gate-green; this reopens when its turn comes.

ndelangen:
Folded into #580 permanently. Three independent signals said this boundary did not exist: main's capture page would not typecheck against the new shared schema, the capture-bundle digest moved with the model rather than the worker, and CI proved it by running the old regression driver against the new snapshot envelope. The stack is three slices now: #580 (model + backend + publisher), #582 (kit), #583 (pages).

Issue #641, opened 2026-08-22
Work the kit compliance wave


ndelangen:
## Question

Execute the kit compliance wave: work the 32 findings on #629 (6 breaks-taxonomy, 15 drift, 11 polish), re-scoped by the Paper decision and the page-h1 amendment this ticket is blocked by. Includes the two kit components missing stories (ConfirmDeleteButton, MemberCountInput) from #633. Small reviewable PRs; the orchestrator reviews before Norbert; visual changes get a browser pass with varied real data.

ndelangen:
Resolved: every finding of the kit sweep (#629) is either merged, decided, or explicitly re-routed. The wave's arc, for the record:

Delivered across eighteen merged PRs: the Paper mapping removal with the Drawer save (#652), the headings amendment and the page-title Block with its header-only guard (#650, #656, #657), the Copperplate convergence across all 47 title sites with the theme alignment (#661, #665, #676), the story and docs batch (#678), six membrane narrowings (#680), the five taxonomy amendments (#681), the IconAction dedupe with declared drag attributes (#684), the chip citation (#686), the ConnectedTabs no-split repair (#689), the WorkbenchLayout settle-handler move under derive-don't-mirror (#691), and CanvasScale fitting in CSS (#705), which closed the last item.

Decided along the way, each on its ticket: Paper (#637), the page h1 (#638), derive-don't-mirror (#639), the widget rung (#640), Copperplate (#662), tone follows the band, PageTitle is header-only, ConnectedTabs stays whole, CanvasScale simplifies rather than earning an exemption (the Firefox evidence comment above).

Re-routed rather than lost: the #451 identity band and the FaqList empty-state sites ride the composition wave (#701); the useValidationHeaderOpen effects ride the declarative wave (#707) with the editors wave's #690; the census carries the residual effect-and-observer findings.

Territory two of the audit map is closed.

PR #661, opened 2026-08-23
Eight page titles that were already Mantine headings become the Block

ndelangen:
3b's mechanical conversions for the kit compliance wave (#641). **Based on `main`, not stacked.**

Deliberately **half** the slice. Eight of the eighteen sites I set out to convert are here; the other ten are held, and the reason is worth reading before reviewing the diff.

## What is here changes nothing on screen

These eight sites spelled their title `Title order={1}`, which is exactly what `PageTitle` renders. Same face, same size, same margins, same position in the header.

| file | sites |
|---|---|
| `routes/_app/[_]_icons.tsx` | 1 |
| `routes/_app/[_]_jobs.tsx` | 1 |
| `routes/_app/groups/$groupSlug/index.tsx` | 3 |
| `routes/_app/factions/create.tsx` | 1 |
| `routes/_app/factions/$factionId/edit.tsx` | 1 |
| `routes/_app/assets/-assetEditorStates.tsx` | 1 |

`assets/-assetEditorStates.tsx` also drops a `Stack` that wrapped the title alone once the Block took over the grouping, the same tidy-up future-plans got.

## Why the other ten are not here

**A bare `<h1>` and a `Title order={1}` are different typefaces in this app, and I did not know that until I measured.**

- `src/app/styles/fonts.css:76` sets `h1` through `h6` to `C_Copperplate_Gothic`, as an element selector.
- `src/app/ui/theme.ts:97` sets Mantine's `headings.fontFamily` to `C_Candara`.
- A class beats an element selector, so every Mantine `Title` ignores the first rule.

Measured, same viewport, both a sole `h1` in a default header:

| page | spelling | face | size | margins |
|---|---|---|---|---|
| `/rulesets` (unconverted) | bare `<h1>` | **`C_Copperplate_Gothic`** | 32px | 21.44px |
| `/profiles` (converted in a draft) | `PageTitle` | **`C_Candara`** | 34px | 0 |

Copperplate is the engraved face that matches the `DUNE` wordmark in the nav. Candara is the body face. So the app has three page-title faces today (Desdemona on hero pages, Candara wherever `Title` was used, Copperplate wherever a bare `h1` was), and **which one a page gets depends only on which spelling its author reached for.** That is precisely the drift `PageTitle` exists to end, so it is a real question rather than a bug to paper over.

Converting the remaining ten would have changed the face on eight files, silently, inside a PR labelled mechanical. Norbert's call was to split here and settle the face on its own merits. Tracked separately; those ten stay on their bare `h1` until it is decided.

## Verified

Browser pass on every changed page, not a diff review: both dev routes, `/factions/create`, the shared asset editor message frame (reached through a missing slug so the real component renders), and `/groups/dreamers`. Each has exactly one `h1`, still centred by its own header column, `C_Candara` at 34px with zero margins, and unchanged sibling gaps because the parent `Stack` still owns them. The alignment-transparency rule from #657 gets its first mass exposure here and holds: every converted title adopts its container's alignment rather than declaring one.

### DIFF
--- a/src/app/routes/_app/[_]_icons.tsx
+++ b/src/app/routes/_app/[_]_icons.tsx
@@ -1,6 +1,7 @@
 import { Box, Chip, Group, SegmentedControl, SimpleGrid, Stack, Tabs, Text, TextInput, Title } from '@mantine/core';
 import { BACKGROUND, DECAL, GENERIC, ICON, LOGO, TROOP, TROOP_MODIFIER } from '@shared/assetIds';
 import { createFileRoute } from '@tanstack/react-router';
+import { PageTitle } from '@ui/block/PageTitle';
 import { TOPIC_ICON_TOPICS, TopicIcon } from '@ui/content/TopicIcon';
 import type { TopicIconTopic } from '@ui/content/TopicIcon';
 import { PageLayout } from '@ui/layout/PageLayout';
@@ -241,7 +242,7 @@ function IconsPage() {
     <PageLayout>
       <PageLayout.Header size="compact">
         <Stack align="center" gap="xs">
-          <Title order={1}>Icon catalog</Title>
+          <PageTitle title="Icon catalog" />
           <Text ta="center" maw={680}>
             Browse the canonical application topics, Lucide library, and Dune SVG assets available in this project.
           </Text>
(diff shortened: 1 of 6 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (one conversion commit plus a main merge): approved pending CI. Verified in the diff that all eight sites were already Title order={1}, zero bare h1 conversions rode along, so face-neutrality holds by construction. The stop-and-split on the typeface discovery was exactly right: ten silent face changes inside a PR labelled mechanical is the failure mode this campaign exists to end, and #662 with measurements is the correct shape for the decision.

Issue #662, opened 2026-08-23
A page title's typeface depends on which spelling its author used


ndelangen:
A page title's typeface currently depends on which of two spellings its author reached for, and nothing says which is correct. This blocks the last ten conversions in 3b of the kit compliance wave (#641).

## The two rules

- `src/app/styles/fonts.css:76` sets `h1` through `h6` to `C_Copperplate_Gothic`, as an element selector.
- `src/app/ui/theme.ts:97` sets Mantine's `headings.fontFamily` to `C_Candara`, the same family as body text.

A class beats an element selector, so every Mantine `Title` ignores the first rule. Both rules look deliberate: neither is a leftover.

## Measured

Same viewport, both a sole `h1` in a default-size header:

| page | spelling | face | size | margins |
| --- | --- | --- | --- | --- |
| `/rulesets` | bare `<h1>` | `C_Copperplate_Gothic` | 32px | 21.44px |
| `/profiles` | `PageTitle` (drafted) | `C_Candara` | 34px | 0 |

Copperplate is the engraved face that matches the `DUNE` wordmark in the navigation. Candara is the body face.

## So the app has three page-title faces

| pages | face | how |
| --- | --- | --- |
| home, future-plans | `C_Desdemona` | `PageTitle` under a `hero` header |
| factions catalogue, `assets/$type`, groups detail, faction create and edit, the two dev routes, the asset editor message frame | `C_Candara` | `Title order={1}`, now `PageTitle` |
| privacy, auth login and error, admin migrations, groups create and edit, rulesets index and create, both FAQ routes, profiles index, profile detail, `AppNotFound` | `C_Copperplate_Gothic` | bare `<h1>` |

Which face a page gets is not a decision anyone made per page. It follows from the spelling, which is exactly the drift `PageTitle` exists to end.

## What needs deciding

What face should a non-hero page title use?

1. **Candara**, the current `PageTitle` behaviour. Ten sites across eight files lose the engraved face. The `fonts.css` heading rule becomes dead for page titles, though it still governs `h2` through `h6`.
2. **Copperplate Gothic**, matching the wordmark and the global rule. `PageTitle` would set the family itself. Wider blast radius: the pages already converted, including the factions catalogue and `assets/$type`, change to Copperplate from what they render today.

Worth deciding alongside it: whether `h2` through `h6` should keep Copperplate either way, since a page whose `h1` is Candara and whose `h2` is Copperplate reads oddly.

## Blocked on this

[...]

Issue #451, opened 2026-08-17
Detail-page identity headers are hand-rolled three times over


ndelangen:
Three detail pages compose the same header band — media in its own column, with a breadcrumb, a title and a meta line stacked beside it sharing one left edge — and each one hand-rolls it with its own markup and its own CSS module classes.

| Page | Composition | Its own CSS |
|---|---|---|
| `factions/$factionId` | `Group` → `FactionToken` + `Stack`(breadcrumb, title, meta) | `.pageHead`, `.factionSymbol`, `.pageHeadText` |
| `rulesets/$rulesetSlug` | `Group` → `Avatar` + `Stack`(breadcrumb, title, meta) | `.pageHead`, `.pageHeadText` |
| `profiles/$profileSlug` | `.identityRow` → `img`/initials + `Stack`(name, hint, summary) | `.identityRow`, `.avatar`, `.avatarPlaceholder`, `.displayName` |

`groups/$groupSlug` currently renders a bare `Title` and would plausibly want the same thing.

## Why it is worth extracting

It has already gone wrong once. The ruleset page's re-cut ([#449](https://github.com/ndelangen/dunezone/pull/449)) put the avatar inline with the title instead of in its own column, which indented the title 40px while the breadcrumb and meta line stayed at the band's left edge. Measured left-edge spread was 71px. Nothing caught it — no test, no lint, no story — because each page owns its own arrangement, so there is nothing for a regression to be a regression *against*.

Three callers is well past the threshold AGENTS.md sets for extraction, and the shape qualifies as a **Layout** rather than a Block: it takes slots and decides where things go, never what they are.

## Directions, none decided

- **A kit Layout** with named compound slots (media, and a text region), container-query responsive like its siblings, living in `src/app/ui/layout`. All three pages adopt it and delete their local classes.
- **Leave it**, and accept that each detail page tunes its own band.

Open questions if it is extracted: whether the text region is one slot or three (breadcrumb / title / meta), and whether the media slot fixes a size or lets callers pass their own — the faction page's token, the ruleset's cover avatar and the profile's round avatar are three different shapes.

Surfaced while working the map in [#426](https://github.com/ndelangen/dunezone/issues/426); out of scope there, since that map's destination is the ruleset page alone and this touches three others.

ndelangen:
**A requirement for the extracted Layout, learned the hard way:** the media should match the height of the text block beside it, and that height cannot be computed in CSS.

[...]

PR #665, opened 2026-08-23
Detail-page placeholder titles become the Block

ndelangen:
3b's second slice for the kit compliance wave (#641). **Based on `main`, not stacked.** Independent of [#661](https://github.com/ndelangen/dunezone/pull/661): no overlapping files, so they can land in either order.

Seven sites, all of them the title a detail page shows while it is loading, erroring, or reporting the thing does not exist.

| file | sites | states |
|---|---|---|
| `routes/_app/factions/$factionId/index.tsx` | 3 | pending, error, no-page |
| `routes/_app/rulesets/$rulesetSlug/index.tsx` | 3 | pending, error, not-found |
| `routes/_app/assets/$type/$slug/index.tsx` | 1 | `AssetDetailMessage`, which serves all three |

All seven were already `Title order={1}`, so **none of them changes typeface.** The bare-`<h1>` sites are held on [#662](https://github.com/ndelangen/dunezone/issues/662).

## These files are half converted on purpose

Each of the three keeps its loaded-state title, because each carries a page-specific class that is blocked on [#451](https://github.com/ndelangen/dunezone/issues/451). `PageTitle` takes only `title` and `eyebrow`, and those classes supply `overflow-wrap: anywhere` and a colour override with nowhere to go.

So a reviewer will see `PageTitle` and `Title order={1}` in the same file. That is the state of the work, not an oversight.

## `rulesets/$rulesetSlug/edit.tsx` is left out entirely

Its falsy arm (`<Title order={1}>Edit ruleset</Title>`) qualifies on every criterion and I still skipped it. Its header is a ternary whose *other* arm is exactly the hand-rolled `.pageHead` markup #451 covers. Converting one arm leaves two spellings inside one expression, and #451 has to rewrite that whole expression anyway, so doing it now saves nobody a visit. It goes with #451.

## Verified by reaching the states, not by reading the diff

Each state was reached deliberately, through slugs that do not exist, so the real components rendered:

| page | title | face | size | centred | back link |
|---|---|---|---|---|---|
| `/factions/no-such-faction` | Faction | `C_Candara` | 34px | yes | 10px below |
| `/rulesets/no-such-ruleset` | Ruleset | `C_Candara` | 34px | yes | 10px below |
| `/assets/card-treachery/no-such-card` | Treachery cards | `C_Candara` | 34px | yes | 10px below |

One `h1` each. The 10px gap is the parent `Stack`'s own `xs`, unchanged, because the Block only owns the space between an eyebrow and the title and these have no eyebrow.

I also checked the loaded states these files keep, to confirm the half conversion left them alone: `/factions/spacing-guild` still renders `.factionTitle` with `overflow-wrap: anywhere`.

## Verified

### DIFF
--- a/src/app/routes/_app/assets/$type/$slug/index.tsx
+++ b/src/app/routes/_app/assets/$type/$slug/index.tsx
@@ -16,6 +16,7 @@ import type { RulesetAssetSlot } from '@shared/rulesets/assetSlots';
 import type { ErrorComponentProps } from '@tanstack/react-router';
 import { createFileRoute, Link, notFound } from '@tanstack/react-router';
 import { OpenableTile } from '@ui/block/OpenableTile';
+import { PageTitle } from '@ui/block/PageTitle';
 import { Section } from '@ui/block/Section';
 import { formatRelativeDate } from '@ui/content/dates';
 import { ProfileLink } from '@ui/content/ProfileLink';
@@ -86,7 +87,7 @@ function AssetDetailMessage({ children }: { children: ReactNode }) {
     <PageLayout>
       <PageLayout.Header>
         <Stack align="center" gap="xs">
-          <Title order={1}>{label}</Title>
+          <PageTitle title={label} />
           <Anchor renderRoot={(rootProps) => <Link {...rootProps} to="/assets/$type" params={{ type }} />}>
             Back to {label.toLowerCase()}
           </Anchor>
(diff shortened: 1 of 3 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (one): approved pending CI. All seven conversions were already Title order={1}, zero bare h1 touched, so face-neutrality holds by construction and nothing pre-empts #662. The deliberate skip at rulesets/$rulesetSlug/edit.tsx:177 is endorsed with its reasoning: converting one arm of a ternary whose other arm is #451's hand-rolled band leaves two spellings inside one expression that #451 rewrites anyway. The self-correction on the ink override (inconsistency, not a contrast failure, measured under forced light scheme) is noted with approval; withdrawing a stronger implication before it hardens into a filed defect is the same discipline as the withdrawn 1.02 figure.

PR #676, opened 2026-08-23
Headings take the engraved face, and the last eleven titles convert

ndelangen:
Implements the decision on [#662](https://github.com/ndelangen/dunezone/issues/662) and closes out 3b's unblocked work. **Based on `main`, not stacked.**

Norbert decided Copperplate Gothic for non-hero page titles, then chose the mechanism after the enumeration: **align the theme** rather than set the family on `PageTitle`.

## Why the theme rather than the Block

`styles/fonts.css:76` has always set `h1` through `h6` to `C_Copperplate_Gothic`, as an element selector. `theme.ts` set Mantine's headings to the body face. A class beats an element selector, so which face a heading rendered in depended only on whether its author wrote `<h2>` or `<Title order={2}>`.

Setting the family on `PageTitle` would have left that alive one level down. The theme was the accident, so the theme is what changes.

## The radius, measured rather than assumed

I checked what `headings.fontFamily` actually reaches instead of saying "every Mantine component":

- Only `Title.css` and `Typography.css` consume `--mantine-font-family-headings`.
- **`Typography` is used zero times in this repo.**

So this reaches the `Title` component and nothing else: 21 page titles through `PageTitle`, 9 direct `order={1}`, and 36 at order 2 and below.

## The eleven conversions ride along because they became face-neutral

With the theme aligned, converting a bare `<h1>` no longer changes its face: those pages already rendered Copperplate. That is why the sites held back from [#661](https://github.com/ndelangen/dunezone/pull/661) are here now.

`auth/error`, `auth/login`, `admin/migrations` (two spellings hoisted to one), `groups/create`, `groups/$groupSlug/edit` (2), `rulesets/create`, `profiles/index`, `profiles/$profileSlug/index` (placeholder only), `AppNotFound`.

They gain 2px and lose the browser's default 21.44px heading margins, both because the Block owns its spacing.

## Two things to look at rather than take on trust

**The small sizes had no precedent.** Every bare heading in the app is `h1` or `h2`, so nothing showed what Copperplate does at `h3`/`h4`. Checked on the faction detail page: 18px and 16px both read clearly. The concern did not materialise.

**Home's story headings grew, and the last commit pins them back.** Norbert's follow-on decision on #662: the face change was about face, not scale, and the extra wrapped lines were a layout regression riding it. The page already owns this size through `storyTitle`, the same species as future-plans keeping its own mobile `h1` tuning.

Copperplate sets wider than Candara at the same size, so the old numbers bought more wrapping rather than bigger type. Every value in the clamp is scaled by **0.815**, the ratio measured at the wrap point (three lines up to 36.5px, four at 37px), so the responsive shape is unchanged and only the face's extra width is paid for. The mobile override is scaled by the same ratio.

Measured at three widths against what the page rendered before the theme moved:

| width | before the theme change | pinned |
|---|---|---|
| 375px | 110px, 3 lines | 90px, 3 lines |
| 1280px | 132px, 3 lines | 107px, 3 lines |
| 1600px | 210px, 4 lines | 214px, 5 lines |

The 1600px row corrects my own earlier framing: the original was **already** four lines at that width, so the regression was narrower than the 219-against-132 figure suggested. Heights now sit at or below the original everywhere.

**One heading is deliberately left alone.** "Your idea belongs at the table" also gained a line, 35px to 70px, but it carries no `storyTitle` class and sits in a different section, so pinning it would be a second page-local decision rather than the one that was made.

## The page that proves the point

`/privacy` rendered a Copperplate `h1` above nine Candara `h2`, on one page, before this. Two heading voices in one document. It now reads as one. That page is also why the Block-only mechanism was the wrong tool: its `h1` was already Copperplate, so setting the family on `PageTitle` would have left it exactly as split.

## Verified

Browser pass across the hero, `/privacy`, the faction detail body at every heading level, and a converted page. The hero still renders `C_Desdemona` at 85.76px, since `PageTitle`'s own stylesheet owns that and outranks the theme.

## What is left of 3b

Five class-carrying titles on [#451](https://github.com/ndelangen/dunezone/issues/451), and five small per-page calls (privacy's kicker, rulesets index, both FAQ routes, auth index). `router.tsx` never converts: it is the `defaultErrorComponent` and covers routes that render outside `MantineProvider`.

### DIFF
--- a/src/app/routes/_app/admin/migrations.tsx
+++ b/src/app/routes/_app/admin/migrations.tsx
@@ -1,5 +1,6 @@
 import { Group, Stack } from '@mantine/core';
 import { createFileRoute, Link } from '@tanstack/react-router';
+import { PageTitle } from '@ui/block/PageTitle';
 import { IconAction } from '@ui/control/IconAction';
 import { PageLayout } from '@ui/layout/PageLayout';
 import { Surface } from '@ui/surface';
@@ -20,6 +21,8 @@ function formatDate(timestamp?: number) {
   return new Date(timestamp).toLocaleString();
 }
 
+const migrationsHeader = <PageTitle title="Migration activity" />;
+
 function AdminMigrationsPage() {
   const loaderData = Route.useLoaderData();
   const profile = useCurrentProfile();
@@ -30,9 +33,7 @@ function AdminMigrationsPage() {
   if (!profile.data?._id) {
     return (
       <PageLayout>
-        <PageLayout.Header>
-          <h1>Migration activity</h1>
-        </PageLayout.Header>
+        <PageLayout.Header>{migrationsHeader}</PageLayout.Header>
         <PageLayout.Content>
           <Surface padding="lg">
             <p>
@@ -46,9 +47,7 @@ function AdminMigrationsPage() {
 
   return (
     <PageLayout>
-      <PageLayout.Header>
-        <h1>Migration activity</h1>
-      </PageLayout.Header>
+      <PageLayout.Header>{migrationsHeader}</PageLayout.Header>
       <PageLayout.Content>
         <Stack gap="sm">
           {dashboardQuery.isPending && <p>Loading migration dashboard…</p>}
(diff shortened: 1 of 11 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (one): approved pending the settling checks. The mechanism enumeration held (the headings variable reaches Title and the unused Typography.css, nothing else), the eleven conversions became face-neutral by riding the same PR as the alignment, and /privacy's one-page Copperplate-h1-over-Candara-h2 split is the concrete proof the theme alignment was the right mechanism over Block-only. The story-heading growth on home is correctly flagged as Norbert's call, not silently pinned.

PR #678, opened 2026-08-23
Four Controls get stories, and nine comments stop describing other code

ndelangen:
Slice K2 of the kit compliance wave (#641), the mechanical stories-and-docs batch. **Based on `main`, not stacked.**

## Ten stories, four files

`ConfirmDeleteButton`, `MemberCountInput`, `AddAction` and `DetachedSurfaceBoundary`. Pure args, no decorators, `fn()` for callbacks, and each lands under an **existing** root rather than inventing one: nine under `Controls`, one under `Surfaces`.

## Nine comments that described something other than the code beside them

Every one of these is a doc block attached to a declaration it does not describe. Sourced from #629's polish list.

| file | what it claimed | what is true |
|---|---|---|
| `block/ProposedContent.tsx` | "lives in the application rather than the interface kit" | the file is in the kit |
| `content/ComplexityGlyph.tsx` | two JSDoc blocks stacked over one function | the first belongs to a record 14 lines below |
| `content/complexity.ts` | module blurb floating over an unrelated threshold | that threshold already had its own doc |
| `control/PreviewChoice.tsx` | component doc separated from the component | it landed on a helper |
| `layout/AsymmetricSplitLayout.tsx` | the only JSDoc sits on the props type | hovering the component showed nothing |
| `layout/WorkbenchLayout.tsx` + its story | explained itself by naming `ConnectedTabs` | a Surface it never mounts, in a Layout that must know nothing of its contents |
| `list/FactionList.stories.tsx` | "one column once the container is narrow" | bottoms out at **two** columns, keyed on `@media` not a container query |

#629's first polish item, the layouts roster in `ui-design-decisions.md`, is **already landed in #643**, so this is ten items rather than eleven.

## The TopicIcon comment, and the gap it was hiding

Its comment credited the whole file to the registry: "a topic added there is tested here without anyone remembering to extend a list."

Only `ALL_TOPICS` is derived, and it carries the weak check (does it render at all). The two hand-written lists carry the sharper assertions, and I checked them against the registry:

- `MASK_TOPICS` lists 10 of 14 masks. Missing all four `complexity*`.
- `COMPONENT_TOPICS` lists 4 of 7 components. Missing `identity`, `about`, `contents`.

So the lists had already drifted, exactly as the comment promised could not happen. The comment now states which check is automatic and which are not, and names today's gap.

**Completing those lists is a real change to test coverage, so it is not in this batch.** Happy to do it as a follow-up; flagging rather than smuggling it into something labelled mechanical.

## One item removes a test, and it is the only non-additive change here

`buildConnectedTabsPath` was exported solely for its test; the component calls it once internally. The test that needed the export asserted fragments of an SVG path string (`toContain('M 8 0')`), which is a helper's internals rather than a contract, and the rule #629 cites is that tests do not dictate API shape.

The helper is private now, with a doc saying why, and that one case is gone. The component's own tests are untouched: rendering, panel switching, disabled semantics. Unit suite goes 656 to 655.

Easily reverted if you would rather keep the geometry pinned; say so and I will restore it through the component's public surface instead.

## Verified

Storybook was run and looked at, not just executed: all ten stories render, `/index.json` confirms they registered under `Controls` and `Surfaces`, and the `DetachedSurfaceBoundary` story's console is clean, which for a component that renders only a context provider is the assertion that matters.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 655 unit tests, 349 storybook tests (up from 339).

`check:prose` caught me writing "simply" in a story doc, which is the guard doing its job on its own author.

### DIFF
--- a/src/app/ui/block/ProposedContent.tsx
+++ b/src/app/ui/block/ProposedContent.tsx
@@ -12,7 +12,6 @@ export interface ProposedContentProps {
  * Marks a shipped region as scaffolding rather than data.
  *
  * The badge is the whole component: without it, prose describing a feature that does not exist reads as a factual statement about the record on screen.
- * It lives in the application rather than the interface kit because it describes our roadmap, not a way of presenting content.
  * Every use of it is a debt to be removed, so `grep ProposedContent` should list them all.
  */
 export function ProposedContent({ label, children }: ProposedContentProps) {
--- a/src/app/ui/content/ComplexityGlyph.tsx
+++ b/src/app/ui/content/ComplexityGlyph.tsx
@@ -11,7 +11,6 @@ const PROGRESS_RING_STROKE = 3;
 const PROGRESS_RING_RADIUS = (PROGRESS_RING_SIZE - PROGRESS_RING_STROKE) / 2;
 const PROGRESS_RING_CIRCUMFERENCE = 2 * Math.PI * PROGRESS_RING_RADIUS;
 
-/** The one place a tier's presentation is defined; every surface reads it from here. */
 /** The x/10 slider positions where each tier's glyph marks the track, shared by every slider. */
 export function complexityTierSliderMarks(size = 12) {
   return [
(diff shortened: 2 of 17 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (two): approved; fully green. Four stories, the hold-state coverage on ConfirmDeleteButton per the delete convention, and nine comment corrections without an AGENTS.md touch. Ready for Norbert.

PR #680, opened 2026-08-23
Six kit components stop accepting more than they mean

ndelangen:
Slice K3 of the kit compliance wave (#641), the membrane-drift batch. **Based on `main`, not stacked.**

Every caller was enumerated before each narrowing, so each is provably lossless rather than argued to be. Per item: what the caller loses, and why nothing breaks.

| component | what a caller could hand it | what it can hand now | why nothing breaks |
|---|---|---|---|
| `block/OpenableTile` | `caption: ReactNode` | `caption: string` | both callers already pass a string |
| `content/ProfileLink` | `children: ReactNode` replacing the name | `label?: string` | one caller used it, passing words |
| `surface/Toolbar` | every `div` prop, via an undeclared spread | `className` and `children` | all 14 call sites are a bare `<Toolbar>` |
| `layout/AtlasLayout` | `Sidebar className` landing on the positioned wrapper | nothing | its one use was the sticky offset, now the layout's |
| `list/FaqList` | nothing; **it** owned the page's empty prose | `emptyLabel`, `noMatchesLabel` | its one caller passes the same strings |
| `control/ListLengthActions` | n/a | n/a | remove side deduped onto `IconAction` |

Plus `FaqList.module.css` dropping a pinned palette step.

## Verified by measurement, not by diff

- **AtlasLayout**: sticky at **24px** before and after. The page's `.atlasSidebar { top: 1.5rem }` is gone and the layout's own rule carries it, so the one page that used it renders identically.
- **FaqList colour**: was `--mantine-color-dark-8`, the same value in both schemes, in an app with a real dark mode. Now `--color-text`, measured flipping between `rgb(46, 41, 39)` and `rgb(236, 231, 220)`.
- **FaqList empty prose**: the ruleset page still shows "No FAQ items yet.", now passed by the page.
- **ListLengthActions**: the remove button measures the same **22×22**, same `aria-label`, same `type="button"`, same red light variant, and is now symmetric with the add button beside it.
- **ProfileLink**: new `Labelled` story renders "Question by Central".

## Two items are held rather than done

**The ConnectedTabs story cast is bigger than #629 reads.** I tried it: removing the cast does not typecheck, and the reason is structural. The story's args **cannot** be the component's props, because `items` is required by `ConnectedTabsProps` and is constructed by the story's own `render` from knobs the component does not have. The cast papers over a real mismatch rather than being a stray annotation.

`ConnectedTabs` is already a K4 item wanting a split proposal ("receiver, producer, and control in one component"), and the story's shape follows whatever that split decides. Doing it now would be work thrown away.

**The SortableReorderHandle dedupe** touches a file [#678](https://github.com/ndelangen/dunezone/pull/678) also touches, and 678 is still open. Held to keep a merge conflict off Norbert's desk, not because it is difficult. It lands as a one-item follow-up once 678 merges.

## One inconsistency found and deliberately not fixed

`ProfileLink` renders `label` as bare text, while the plain username goes through `.username`, which carries `font-weight: 600` and `color: var(--color-text)`. So a caller handing it words gets no name treatment.

That predates this change, since the `children` slot rendered bare too, and it is arguably wrong now that `label` is data rather than a slot. I left it because correcting it would restyle the one page that uses it, and this deployment has no FAQ data to render that page, so I could not look at the result. Worth a decision rather than a guess.

## Verified

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 656 unit tests, 340 storybook tests.

### DIFF
--- a/src/app/routes/_app/future-plans/index.module.css
+++ b/src/app/routes/_app/future-plans/index.module.css
@@ -13,10 +13,6 @@
   padding: 3.5rem clamp(0rem, 2vw, 1.5rem) 1rem;
 }
 
-.atlasSidebar {
-  top: 1.5rem;
-}
-
 .indexLink {
   padding: 0.9rem 0;
   color: inherit;
--- a/src/app/routes/_app/future-plans/index.tsx
+++ b/src/app/routes/_app/future-plans/index.tsx
@@ -148,7 +148,7 @@ function FuturePlansPage() {
       </PageLayout.Header>
       <PageLayout.Content>
         <AtlasLayout className={styles.atlasFrame}>
-          <AtlasLayout.Sidebar className={styles.atlasSidebar}>
+          <AtlasLayout.Sidebar>
             <Stack component="aside" gap="md">
               <Badge color="gray" variant="filled" w="fit-content">
                 All planned
--- a/src/app/routes/_app/profiles/$profileSlug/index.tsx
+++ b/src/app/routes/_app/profiles/$profileSlug/index.tsx
@@ -70,9 +70,8 @@ function AskerChip({
       username={profile.username}
       avatar_url={profile.avatar_url}
       className={styles.askerLink}
-    >
-      Question by {profile.username ?? 'Unknown'}
-    </ProfileLink>
+      label={`Question by ${profile.username ?? 'Unknown'}`}
+    />
   );
 }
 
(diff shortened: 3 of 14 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (one): approved pending the two settling checks. All six narrowings verified against the report: caption and label to strings with the caller enumerations in the body, the Toolbar spread dropped against fourteen bare call sites, the sidebar offset absorbed into the layout that owns placement, FaqList's empty words handed back to the page as data, the color pin now scheme-flipping with measurements, and the remove side deduped onto IconAction. Holding the ConnectedTabs story cast for the K4 split proposal is right and the attempted-first evidence (the cast covers a real args-versus-props mismatch) upgrades it from deferral to dependency. The ProfileLink label treatment question goes to Norbert.

PR #682, opened 2026-08-23
The animated leader token moves to the one page that renders it

ndelangen:
K4's first item from the kit compliance wave (#641), and the only one in that slice that is a plain rule application rather than a design question. **Based on `main`, not stacked.**

## Why it was never Content

[#629](https://github.com/ndelangen/dunezone/issues/629) put it plainly: *kind is judged at the membrane*, and `AnimatedLeaderToken` had no membrane. No props at all. `PORTRAITS` and `EDITS` were both baked in. One importer, the home page.

There was nothing crossing the boundary for a category to be judged at, so the question was never "which kit category is this" but "why is this in the kit".

## Where it goes

Into `routes/_app/index.tsx`, which already keeps `HomepagePending`, `HomepageError` and a formatting helper. A single-caller view component belongs in its caller's file; `-` files are for logic modules.

The stylesheet moves into `index.module.css`, with names prefixed (`leaderToken`, `leaderTokenPrevious`, …) and the keyframes renamed, because that file carries the whole page rather than one component.

**The story goes with it.** A route-local component has no story in this repo, and the rule that every kit component has one applies to the kit it has now left. Storybook drops 340 to 339; stylesheet count drops 63 to 62.

## Verified against the running page

Same 320px circle, same `1 / 1` aspect ratio, same 50% radius, same drop shadow, same accessible name, rendering the same leader and strength.

**The animation took some working for, and the reason is worth recording.** It did not animate on first check. That is not a regression: the browser reports `prefers-reduced-motion: reduce`, and the component deliberately returns early from its timer effect under it. A correctly-frozen animation and a broken timer look identical from outside.

Patching the media query and remounting showed the state machine intact: it had already cycled off the first leader and was typing the next name a letter at a time, `F`, `Fa`, `Far`, `Faro`, `Farok`.

## Verified

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 656 unit tests, 339 storybook tests.

## What is left of K4

`CanvasScale` and `ConnectedTabs` get written proposals before any code. `WorkbenchLayout` is unblocked now that [#639](https://github.com/ndelangen/dunezone/issues/639) decided derive-don't-mirror, and its `onBlurCapture` threading gets judged against that rule. The `ConnectedTabs` story cast from [#680](https://github.com/ndelangen/dunezone/pull/680) folds into its proposal, since the story's args cannot match the component's props while `items` is constructed by the story's own render.

## SonarCloud, and what I did about it

The quality gate failed on this PR with a **C Security Rating on New Code**. Both findings are in code this PR *moved* rather than wrote, and the API names them precisely:

| rule | line | finding |
|---|---|---|
| `typescript:S3358` | the phase ternary | nested ternary picking the displayed name |
| `typescript:S2245` | the portrait pick | `Math.random`, typed VULNERABILITY, tagged `former-hotspot` |

**S3358 is fixed**, as a `switch`. That is what this repo reaches for over a ternary chain in multi-branch code, so the gate and the house style want the same thing here.

**S2245 is left alone, deliberately.** It picks which of three portraits the token wears. It is choosing an illustration, not a token, a key or an identifier, and swapping in `crypto.getRandomValues` for a decorative image would be worse code that reads as though it were guarding something. `Math.random` already appears five times in `widgets/background-composer/backgroundRandomizer.ts`; those lines are invisible to the gate only because they are not new.

Sonar's own tag is the giveaway: `former-hotspot`. A line that has existed for months scores as a vulnerability the moment it changes file. **This PR introduces no new behaviour and no new randomness.**

If the rating needs to be green rather than explained, the resolution is marking it Safe in SonarCloud, which needs project-admin access I do not have. Flagging rather than working around it.


ndelangen (commented):
Orchestrator review, commits enumerated (two): approved on content; the merge decision rides on the SonarCloud question below. Verified: the one added Math.random line is the relocation of the line main already carries in this same file, backgroundRandomizer.ts holds five identical calls that score nothing, and the PR adds no behaviour and no new randomness. The S3358 fix as a switch matches house style. Declining to swap in crypto randomness for a decorative portrait is endorsed: security-flavoured code that guards nothing misleads the next reader, and the honest resolutions are marking S2245 Safe (project admin) or accepting the red.

PR #683, opened 2026-08-23
The listing face schemas parse what they hand the renderers

ndelangen:
## Summary

- derive the listing face schemas in `AssetFace` from `src/shared/assets/schema` instead of restating them with weakened fields
- remove seven `as never` casts on renderer props, plus one `as BundleBandData` of the same shape
- keep every documented leniency: the objects stay loose, and the optional fields stay optional

Each of the eight casts hid a real type error, not generic plumbing. The local schemas declared `background: z.unknown()`, `image: z.string()`, `decals: z.array(z.unknown())` and `texts: z.array(z.unknown())`, then asserted them back into renderer props typed as `Background`, the asset-id enum, `PlacedDecal[]` and `PlacedText[]`. Nothing had checked the values the renderers were handed, which is what the neutral-face fallback exists to prevent.

The shape authority already covers this. `parseAssetDataForWrite` parses every asset write through the same shared Zod the renderers draw from, so `background`, `image`, `decals` and `texts` are guaranteed at the write seam. The widget was weakening fields the server had already validated and then asserting them back.

No new shared contract was needed. `src/game/data/objects.ts` re-exports `@shared/assets/schema`, so the renderers and the widget were already reading one authority and only the widget's copy had drifted. This follows ADR-0001: types derive from their owner, and the compiler enforces what the casts had silenced.

`TreacheryAsset` now imports from `@shared/assets/schema` alongside the rest, so the file reaches that module by one path rather than two.

## Validation

Passed:

- bun run typecheck
- bun run lint
- bun run format:check
- bun run check
- bun run knip
- bun run test (656 tests, 107 files)
- bun run check:prose, check:css-orphans, check:app-layout

`publisher:release:verify` was not run. `src/app/widgets/**` is not a renderer-manifest input path (`src/shared/`, `src/game/`, `src/app/print/`, `workers/publisher/`, `media/`), nothing under `src/app/print/` imports `AssetFace`, and `publisher:manifest:verify-changed` reports "no capture-bundle changes since real-origin/main".

Behaviour was checked against real data rather than inferred:

- ran the old and the new schemas over all 27 asset rows in the dev deployment, including the `presentedData` join that resolves reference-mode deck cardbacks. No row changes outcome, and every row the widget actually receives parses under the new schemas. The two rows that fail raw are the reference decks the join rewrites before the widget sees them.
- browser pass on a dev server started in this worktree (listener cwd confirmed): `/assets`, `/assets/deck`, `/assets/token-enhance`, `/assets/token-disc`, `/assets/bundle`, and a tech-token detail page showing both faces. Cardback backgrounds and images, rectangle decals and placed texts, bundle bands, and the custom-back path all draw. No console errors.


src/app/widgets/asset-face/AssetFace.tsx
```diff
@@ -189,31 +201,25 @@ function danglingDeckCardback(data: unknown): boolean {
...
+ */
 const cardbackFaceSchema = z.object({
-  cardback: z.looseObject({
-    name: z.string(),
-    background: z.unknown(),
-    image: z.string(),
-    imageScale: z.number(),
-    imageOffset: z.tuple([z.number(), z.number()]).optional(),
-  }),
+  cardback: z.looseObject(CardBackContract.partial({ imageOffset: true }).shape),
```
ndelangen: **The deck editor's live proof blanks on every keystroke while an author types a colour.**

`DeckEditor.tsx:85` renders `CardbackProof` as `<AssetFace type="deck" data={{ cardback }} />` against the live draft, at `:339`, `:362` and `:452`. Mantine's `ColorInput` is controlled and its input handler calls `setValue(rawInputValue)` unconditionally per keystroke, repairing only on blur (`@mantine/core/esm/components/ColorInput/ColorInput.mjs:140-144`); `onChangeEnd` is the one gated on `isColorValid`. `BackgroundComposer.tsx:321` wires `onChange`, not `onChangeEnd`, and passes the raw string up into the draft.

`HEXCOLOR` is `/^#[0-9a-f]{6}$/i`, so every intermediate value fails. Typing `#4B4C0D` one character at a time, same row, same probe:

```
                    main       this PR
typed "#"           DRAWS      neutral
typed "#4"          DRAWS      neutral
typed "#4B"         DRAWS      neutral
typed "#4B4"        DRAWS      neutral
typed "#4B4C"       DRAWS      neutral
typed "#4B4C0"      DRAWS      neutral
typed "#4B4C0D"     DRAWS      DRAWS
```

Whose fault this is matters for the fix. The draft was always transiently invalid; the old schema just never looked. `CardbackData` is `z.infer<typeof CardBack>`, and a regex-refined string infers as plain `string`, so the type guarantees nothing about the value. I went looking expecting to clear this path on exactly that reasoning and it does not hold. This PR is what makes the gap visible, which is the PR doing its job.

So the fix I would prefer is at the source: commit through `onChangeEnd` in `BackgroundComposer`, which fires only when the colour parses. That is arguably its own PR. If you would rather not widen scope, relax `background` for the deck listing schema alone, at the cost of the PR's best field. Either is fine. Shipping the flicker without deciding is the part I would not do.

`DeckEditor` is the only editor exposed: `BundleEditor` draws its draft band through `BundleContainer` directly, and `TokenEditor` and `RectangleTokenEditor` call `CustomToken` and `RectangleToken` directly, all three bypassing this parse.

Storybook cannot catch it and that is not a gap in your testing. `verify / storybook` passes in CI and the cardback stories render committed, valid values; no story types a partial one into a composer.

src/app/widgets/asset-face/AssetFace.tsx
```diff
@@ -189,31 +201,25 @@ function danglingDeckCardback(data: unknown): boolean {
   return typeof data === 'object' && data !== null && 'cardback' in data && data.cardback === null;
 }
 
+/*
+ * A deck's cardback, at the contract the renderer draws.
+ * `imageOffset` is the one relaxation: the call site defaults it, so a row stored before the field existed draws
+ * centred rather than falling to the neutral face.
+ */
 const cardbackFaceSchema = z.object({
```
ndelangen, line 216: **`X.partial({...}).shape` is correct today and has a silent failure mode. Prefer `.loose()`.**

Rebuilding through `.shape` copies the field map and nothing else. Object-level checks do not survive, and Zod 4 will not tell you, because a refined object still exposes `.shape`:

```js
const Refined = z.strictObject({ a: z.string(), b: z.number() }).refine((v) => v.b > 10);
Refined.safeParse({ a: 'x', b: 1 }).success              // false
z.looseObject(Refined.shape).safeParse({ a: 'x', b: 1 }) // true  <- check dropped, no error
```

None of the four schemas carries an object-level refinement today, so this is latent rather than live. But the day someone adds one to `CardBack` (say, imageScale must be 1 when imageOffset is set), the listing quietly stops enforcing it and nothing fails. That is the same second-source failure this PR exists to remove, reintroduced one level up.

Zod 4.4.3 has `.loose()`, which changes only the unknown-key policy and preserves everything else:

```ts
const cardbackFaceSchema = z.object({
  cardback: CardBackContract.partial({ imageOffset: true }).loose(),
});
```

Verified in this worktree: `.loose()` exists on all four and accepts unknown keys exactly as `z.looseObject` does. It also reads closer to what the comment says the schema is, "the stored one with fields relaxed", since the stored one is now literally the subject of the expression.

src/app/widgets/asset-face/AssetFace.tsx
```diff
@@ -189,31 +201,25 @@ function danglingDeckCardback(data: unknown): boolean {
   return typeof data === 'object' && data !== null && 'cardback' in data && data.cardback === null;
 }
 
+/*
+ * A deck's cardback, at the contract the renderer draws.
+ * `imageOffset` is the one relaxation: the call site defaults it, so a row stored before the field existed draws
```
ndelangen: **"`imageOffset` is the one relaxation" understates it, and the second relaxation is load-bearing.**

The wrapper is `looseObject`, and that looseness is what carries `mode: 'custom'`. `DeckAsset.cardback` is a union whose authored member is `CardBack.extend({ mode: z.literal('custom') })`, and `assets_deck_cardback_wrap_v1` rewrote every stored cardback into that shape. `CardBack.shape` has no `mode`, so a wrapped row parses here only because unknown keys are tolerated.

That makes the sentence a trap rather than an inaccuracy. A reader who believes `imageOffset` is the only relaxation and tightens the wrapper to `z.object` breaks every migrated deck at once, and the failure is a neutral face rather than an error, so it would not be obvious what happened.

Naming both relaxations, and naming `mode` as the reason for the second, is a one-line fix.

src/app/widgets/asset-face/AssetFace.tsx
```diff
@@ -189,31 +201,25 @@ function danglingDeckCardback(data: unknown): boolean {
...
-  bottomSecond: z.string().optional(),
-  ring: z.boolean().optional(),
-  ringShadow: z.boolean().optional(),
-});
+const drawableTokenFace = z.looseObject(
+  TokenFace.partial({ symbolScale: true, top: true, bottomFirst: true, bottomSecond: true, ring: true }).shape
+);
 
 const tokenFaceSchema = z.object({
   front: drawableTokenFace,
```
ndelangen, line 236: **A failing back face blanks the front too, on every surface.**

`tokenFaceSchema` puts `front` and `back` in one object, so `safeParse` is all-or-nothing across both. A row whose back names a glyph id that has left the generated `ALL` enum fails the whole parse, `faceForSide(undefined, 'front')` returns undefined, and `<AssetFace side="front">` draws the neutral box even though the front face is perfectly valid.

On main the same row drew its front normally, because `image: z.string()` and `background: z.unknown()` accepted the back too.

Minor, because the trigger is the same generated-enum-moves event as the other two and nothing hits it today. Worth knowing because it doubles the blast radius of that event: a bad back costs the front, on browse tiles, picker rows, bundle member previews and the detail page.

If you keep the parse whole-row, this is fine as long as it is deliberate. If you would rather it degrade, parsing `front` and `back` independently is a small change and keeps a good front face drawing.

ndelangen (commented):
## Review

The direction is right and this should land. It removes a real second source: the local schemas restated the stored shape with the pass-through fields weakened, then asserted them back into typed props, so nothing had checked the values the renderers were handed. Deriving from `@shared/assets/schema` makes the compiler enforce what the casts silenced, which is ADR-0001 applied exactly where it was being dodged. Your evidence gathering was thorough and I did not redo the production-row check; running it through the public `assets:listByTypes` query was the right answer to the dev-replica objection, and the different distribution makes it a real second sample rather than a repeat.

One finding I would fix before merge, three smaller ones, and answers to your four questions.

### The one to fix: the deck editor's live proof blanks while you type a colour

This is the case you asked for in question 1, and it is not hypothetical. It happens on every colour edit today.

`DeckEditor` renders `CardbackProof` as `<AssetFace type="deck" data={{ cardback }} />` against the **live draft**. Mantine's `ColorInput` is controlled and its input handler calls `setValue(rawInputValue)` unconditionally on every keystroke, repairing only on blur (`ColorInput.mjs:140-144`). `BackgroundComposer` passes that raw string straight through (`BackgroundComposer.tsx:321`, and the gradient stop at `:432`) into the draft. So a half-typed hex reaches the parse, and `HEXCOLOR` is `/^#[0-9a-f]{6}$/i`.

Same row, same probe, typing `#4B4C0D` one character at a time:

```
                    main       this PR
typed "#"           DRAWS      neutral
typed "#4"          DRAWS      neutral
typed "#4B"         DRAWS      neutral
typed "#4B4"        DRAWS      neutral
typed "#4B4C"       DRAWS      neutral
typed "#4B4C0"      DRAWS      neutral
typed "#4B4C0D"     DRAWS      DRAWS
```

The proof flickers to a grey initials box for six of the seven intermediate values and returns on the last character.

I want to be precise about whose fault this is, because it changes the fix. The draft was **always** transiently invalid; the old schema simply never looked. `CardbackData` is `z.infer<typeof CardBack>`, and `HEXCOLOR` is a regex-refined string, so its inferred type is plain `string` and the type guarantees nothing about the value. I had assumed the draft was compile-time safe and checked it expecting to clear the path; it is not, and this PR is what makes that visible. That is the PR working as intended, not misbehaving.

So the fix I would prefer is at the source: have the composer commit through Mantine's `onChangeEnd`, which fires only when `isColorValid` passes, rather than `onChange`. That is a separate change and arguably its own PR. If you would rather not widen the scope, the alternative is to keep `background` relaxed for the deck listing schema only, which costs the PR its best field. Either is defensible; shipping the flicker without deciding is the one I would not do.

Scope, since I checked all four editors: `DeckEditor` is the only one exposed. `BundleEditor` draws its draft band with `BundleContainer` directly, and `TokenEditor` and `RectangleTokenEditor` call `CustomToken` and `RectangleToken` directly. All three bypass the parse entirely.

Storybook does not cover this, and that is not a gap in your testing: `verify / storybook` passes in CI, and the cardback stories render committed, valid cardbacks. No story types a partial value into a composer, so the transient-draft path has no coverage by construction.

### Answering your questions

**1. Any case where degrading to neutral is worse?** The keystroke one above, plus two measured trades I would call your judgement rather than defects.

`z.array(PlacedDecal)` is all-or-nothing, so one bad element costs the whole face. `DecalLayer` reads only known fields and turns `decal.id` into an `xlinkHref`, so an unresolvable id draws nothing while everything around it draws:

```
                              main       this PR
two good decals               RENDERED   RENDERED
one good, one stale id        RENDERED   NEUTRAL     <- background, ring, texts and the good decal all lost
```

The same shape applies to `image`: on main a retired glyph id kept the background, ring and curved labels and lost only the symbol.

The counterargument is good and I want it on the record rather than buried: `card-treachery` already parses through the fully strict `TreacheryAsset` on main, so whole-face rejection is an established contract here, not a new policy this PR invents. And the trigger for both is a generated enum moving under stored data, which is a migration-shaped event either way. If you are comfortable with that, say so in the comment and I am satisfied. What I would not leave is the current state, where the file's stated rule ("what the renderer cannot default") does not predict the behaviour, since the renderer defaults nearly all of it.

**2. `X.partial({...}).shape` inside `z.looseObject(...)`: correct today, with a silent failure mode.** Use `.loose()`. Details and proof in the thread.

**3. The `CardBack as CardBackContract` alias: keep it, do not churn.** The renderer appears in JSX in the render tree, which is where a reader reconstructs what the file draws. The schema appears once, in a const declaration. Renaming the thing that shows up in the output to protect the name of the thing that shows up in a type is the wrong trade.

**4. Things the casts were load-bearing for.** I traced every field of the four derivations against its call site and the mask discipline holds in both directions. `symbolScale` is defaulted by `tokenSymbolSize` (`?? 1`), `top` by `face.top || undefined`, the two bottom lines through `tokenBottom`, `ring` by `?? false` and `?? shape === 'round'`, `imageOffset` by `?? [0, 0]`, `decals` and `texts` by `?? []`. Nothing a mask relaxes reaches a renderer undefined, and nothing the call site defaults is missing from a mask. `ringShadow` is already optional upstream in both `TokenFace` and `RectangleTokenFace`, so its absence from the masks is correct rather than an oversight.

`BundleBandData` was already `z.infer<typeof BundleBand>` on main, which is why the eighth cast was purely the local schema's fault and why removing it was safe.

One thing that no check here can cover: all of this proves today's rows pass today's `ALL`. Whether the enum ever moves under stored data is a process question about treating asset removal as a migration, and it is out of scope for this PR.

src/app/widgets/asset-face/AssetFace.tsx
ndelangen: Confirmed and fixed in `dfbc4763eb2`. I reproduced your table exactly before changing anything: six of the seven intermediates of `#4B4C0D` regress.

```
  "#"        main DRAWS     this PR neutral   <-- REGRESSION
  "#4"       main DRAWS     this PR neutral   <-- REGRESSION
  "#4B"      main DRAWS     this PR neutral   <-- REGRESSION
  "#4B4"     main DRAWS     this PR neutral   <-- REGRESSION
  "#4B4C"    main DRAWS     this PR neutral   <-- REGRESSION
  "#4B4C0"   main DRAWS     this PR neutral   <-- REGRESSION
  "#4B4C0D"  main DRAWS     this PR DRAWS
```

You are also right that my earlier "the draft is typed, so it is safe" reasoning was wrong, and for the reason you give: `HEXCOLOR` is `z.string().regex(...)`, which infers as plain `string`, so `CardbackData` guarantees nothing about the value.

I took neither of your two options. Fixing `onChangeEnd` at the source looked right until I read the call site: `BackgroundComposer.tsx:321` is a *controlled* `ColorInput` with `value={value}`, so committing only on `onChangeEnd` stops echoing the value back and the field no longer shows what you type. That fix needs local text state in the composer, it touches every editor that composes a background, and it is a bigger and riskier change than this PR should carry. Relaxing `background` for the deck listing costs the best field, as you say.

The third option is the one the codebase already points at. `TokenProof` and `RectangleProof` both draw the renderer inside the catalogue's frame and never touch the parse; `CardbackProof` was the only editor proof routing a live draft through `AssetFace`. So it now does what its two siblings do:

```tsx
<CardFrame width={width}>
  <CardBack {...cardback} />
</CardFrame>
```

`CardFrame` is exported alongside `TokenFrame`, which the token editors already use for exactly this. That removes the parse from the editor path entirely rather than weakening it, so no keystroke can produce a neutral face.

I checked your scope claim rather than assuming it. Every remaining `<AssetFace>` call site takes `entry.data`, `member.data`, `card.data` or `picked.data`, all stored server reads. No live draft reaches `AssetFace` any more.

src/app/widgets/asset-face/AssetFace.tsx
ndelangen: Adopted in `dfbc4763eb2`. I probed it rather than taking it on trust, and it behaves exactly as you describe on zod 4.4.3:

```
CLAIM 2: does z.looseObject(X.shape) drop an object-level refine?
  refined schema itself       : rejects
  refined object exposes .shape: true
  z.looseObject(Refined.shape): ACCEPTS (refine LOST)

CLAIM 2b: .loose()
  typeof base.loose                          : function
  base.loose() keeps extra keys              : true
  base.partial({b:true}).loose() drops required b: true
```

So the rebuild really does discard an object-level check while still typechecking, which is the second-source failure this PR exists to remove, reintroduced one level down. All four schemas now relax in place:

```ts
band: BundleBand.loose()
cardback: CardBackContract.partial({ imageOffset: true }).loose()
TokenFace.partial({ ... }).loose()
RectangleTokenFace.partial({ ring: true, decals: true, texts: true }).loose()
```

Reran the old-versus-new comparison over production and dev rows after the change to be sure the spelling swap was inert: 0 rows changed outcome on either.

src/app/widgets/asset-face/AssetFace.tsx
ndelangen: Fair, and the failure you describe is the dangerous kind: silent, migration-wide, and with no type error to catch it. Reworded in `dfbc4763eb2` to name both relaxations and say which one is load-bearing.

```
 * A deck's cardback, at the contract the renderer draws, with two relaxations rather than one.
 * `imageOffset` becomes optional because the call site defaults it, so a row stored before the field existed draws
 * centred rather than falling to the neutral face.
 * `loose()` is the second and the load-bearing one: `assets_deck_cardback_wrap_v1` tags an authored cardback
 * `mode: 'custom'`, and `presentedData` passes a non-reference deck through untouched, so the extra key arrives here.
 * Tightening this wrapper back to strict would turn every migrated deck into a neutral face without a type error.
```

src/app/widgets/asset-face/AssetFace.tsx
ndelangen: Fixed rather than deferred, in `dfbc4763eb2`. You filed it as minor, but it doubles the blast radius of the exact failure mode this PR makes reachable, and the whole point of the change is that a face should blank only when the thing it draws is unreadable. A token whose front is perfectly good drawing nothing because its back names a deleted vector is the same defect one level up.

Both back unions now end `.optional().catch(undefined)`, so an unreadable back reads as no back, which is already what a referenced back does. Only the back side falls to neutral.

Verified against a token with a good front and a back naming a vector no longer in the generated enum:

```
malformed back, front is fine:
  previous commit : WHOLE TOKEN FAILS, front blanks too
  current         : parses, front draws
  back resolves to: undefined (back side neutral, front intact)
```

This is a schema-level statement about blast radius, not a runtime guard papering over a type mismatch: the inferred type is still the union or `undefined`, and `faceForSide` already treats an absent back as "nothing to draw here".

Also, thanks for the note on the local storybook failures. Worth recording that they were your worktree: `bun run storybook:test` passes here on this branch, 85 story files and 339 tests, including all four `DeckEditor` stories. Nothing to chase.

ndelangen (commented):
## Re-reviewed at dfbc4763eb2. All four addressed, and the direction taken is better than either option I offered.

I reproduced the three behavioural claims rather than reading the diff and agreeing.

**The `CardFrame` direction is the right call, and not just pragmatically.** `AssetFace`'s contract is "given unknown stored data, draw a face or fall back to neutral". An editor draft is not unknown data; it is a typed value the editor owns. Routing it through the listing parser applied an untrusted-input policy to trusted input, and the neutral fallback that is correct for a listing became a flicker that is wrong for a proof. So this removes a misuse rather than working around a bug, which is why it beats the option I offered of relaxing `background` for the deck schema. That one would have degraded the untrusted-input policy to fix a trusted-input path: the wrong lever, and it would have cost the PR its best field.

Your objection to my other option is correct and I had not thought it through. `BackgroundComposer.tsx:321` is a controlled `ColorInput` with `value={value}`, so committing only on `onChangeEnd` stops echoing keystrokes and the field stops showing what you type. That needs local text state in the composer and reaches every editor that composes a background. Out of scope here, agreed.

**On the interactive confirmation you offered: not needed, and here is why rather than a judgement call.** I rendered the same valid cardback both ways and compared the markup, normalising only the per-mount id prefixes that `useCountId` generates:

```
a) identical markup: true
```

Byte-identical. The swap is visually inert for every valid draft, so there is nothing a browser would show that this does not. Do not go fighting the auth wall for it.

**The flicker is closed at the path, not just structurally:**

```
b) typed "#"        -> DRAWS
b) typed "#4B4"     -> DRAWS
b) typed "#4B4C0D"  -> DRAWS
```

**`catch(undefined)` does exactly what the comment claims, including not over-swallowing:**

```
c) front side, bad back  -> DRAWS      (was: neutral)
c) back side,  bad back  -> neutral    (correct: only the back is unreadable)
c) front side, good back -> DRAWS
c) front side, BAD FRONT -> neutral    (correct: the catch did not swallow a front failure)
```

That last line is the one I most wanted to see, since a `catch` in the wrong place would have quietly turned an unreadable front into a drawn face.

**`.loose()`**: correct, and re-running the row comparison to confirm the spelling swap was inert was the right instinct. A schema change that "should be" behaviour-preserving is exactly the kind that is worth proving rather than asserting.

Every check on this commit is green, 25 passing with Sourcery skipping.

Nothing further from me. Good to merge from my side, and I am leaving that to Norbert.

One observation, not a request. `TokenProof`, `RectangleProof` and now `CardbackProof` are each "frame plus renderer" in three different widgets, and both frames are now exported for that purpose. Three call sites across three widgets is under the bar AGENTS.md sets for a widget, let alone a kit component, so I would leave it. Worth knowing it is the shape that earns one if a fourth appears.

Separately, and genuinely out of scope: the composer writing values into a typed draft that do not satisfy the schema that type claims to represent is still true after this PR. Nothing consumes it now, so nothing breaks. It will bite whatever consumes a draft background next, and it would make a reasonable small ticket.

src/app/widgets/deck-editor/DeckEditor.test.tsx
```diff
@@ -97,3 +106,55 @@ it('keeps a composed cardback across a trip through the reference tile', () => {
...
+      </MantineProvider>
+    );
+    const count = container.querySelectorAll(`img[src="${emblem}"]`).length;
+    unmount();
+    return count;
+  };
+
+  const complete = emblemCount(withFirstColor('#4B4C0D'));
+  /* The face has to be drawing something for the comparison below to mean anything. */
+  expect(complete).toBeGreaterThan(0);
```
ndelangen: **The guard on this line does not guard the thing it was written for.**

I verified the new test the way you did, and it does gate: reverting `CardbackProof` to the `AssetFace` call fails it with `expected 1 to be 4`. Good test, and catching the vacuous version yourself was the right instinct.

But I also ran it with the old `observe() {}` stub restored, and it still passes. Measured counts:

```
dead stub    complete=1  partial=1   -> toBeGreaterThan(0) passes, equality passes, vacuous
firing stub  complete=4  partial=4   -> real
```

The emblem appears once outside `CanvasScale`, in a picker tile, so `complete` is 1 rather than 0 when nothing in the canvas mounts. `toBeGreaterThan(0)` is satisfied by that tile alone, and every prefix agrees at 1, so the comparison holds and the test reports success against a DOM with no proof in it. That is the same shape as the bug you just fixed, one level down: the line that reads as the safety net is not the thing providing safety. What actually makes this test real is the firing observer, and nothing in the file fails if someone simplifies that stub back.

`toBeGreaterThan(1)` closes it, since 1 is exactly the tile's own contribution and the proof supplies three more. Worth saying so in the comment, something like: the picker tile draws this emblem on its own, so a count of one means the canvas mounted nothing.

Not blocking. The committed test does gate today, and the stub carries a comment explaining why it fires.

**Your general warning checks out, and it is live in two sibling files.** `BundleEditor.test.tsx:24` and `TokenEditor.test.tsx:24` both still have `observe() {}`, and both editors render through `CanvasScale`. Neither is vacuous today, because their assertions are on a select's value, a label field's presence and draft state, all outside the canvas. They are one proof assertion away from the trap. Since the fix is four lines and you have it written, copying the firing stub into both now would close the class rather than leave it for whoever adds that assertion.

ndelangen: Confirmed and fixed in `9394181bb71`. I reran the committed test with the old stub restored before changing anything, and got exactly your numbers:

```
complete = 1  partial(#4B4) = 1     -> passes, vacuous
```

So `toBeGreaterThan(0)` was satisfied by the picker tile alone, and you are right that this is the same shape as the bug one level down: the line that reads as the safety net was not the thing providing safety.

Now `toBeGreaterThan(1)`, with the reason in the comment rather than left as a bare number, since the number is only meaningful if you know where the extra emblem comes from:

```
   * More than one, not more than zero, and the difference is the whole guard.
   * The stock picker tile draws this same emblem once, outside the canvas, so a count of exactly one means
   * `CanvasScale` mounted nothing and there is no proof in the DOM to compare prefixes against.
   * `CardBack` draws the emblem three times, so a proof that really mounted puts the count at four.
```

Verified it gates in both directions: dead stub gives `AssertionError: expected 1 to be greater than 1`, firing stub passes.

`BundleEditor.test.tsx` and `TokenEditor.test.tsx` both have the firing stub now, so the class is closed rather than left for whoever adds the first proof assertion. I checked the rest of the tree rather than stopping at your two: eight test files have a dead `observe()`, but the other six (`ConnectedTabs`, `AssignPopover`, `AppRoot`, `ProfilePicker`, profiles `-edit` and `-delete`) test components that never mount `CanvasScale`, so their stubs are correct as they are and I left them alone.

Separately, this branch was `BEHIND` and is now merged up to `f89afd55e03`, which pulled in the kit wave (#678, #680, #681) and the leader token move (#682). Full suite green on the merge, `AssetFace` took changes from main and merged clean, no `as never` reintroduced, and the publisher manifest guard still reports no capture-bundle changes.

PR #689, opened 2026-08-23
ConnectedTabs drops two dead props and its story stops inventing a type

ndelangen:
Closes the `ConnectedTabs` item on [#629](https://github.com/ndelangen/dunezone/issues/629). **Based on `main`, not stacked.** Net **44 lines removed**.

#629 wanted this split three ways: receiver, producer and control. I proposed against it after reading the component, its 395-line stylesheet, all seven call sites, and its story and test. Norbert's call was **no split, repair in place**.

## Why the three-way claim does not hold

**The producer leg is wrong.** It rests on `label: string` being rendered as text (`ConnectedTabs.tsx:13`, rendered `:329`, `:348`, `:379`). By that test half of `ui/control` is misfiled: `PreviewChoice` takes per-option captions, renders them, and is filed as a Control. So is `Select`. An option's name is not produced content.

**Receiver and control are both real, and a split does not remove the overlap.** Radix forces a compound shape (`Tabs.List` and `Tabs.Content` both read `Tabs.Root` context), and a compound root still takes `value`/`onValueChange` and still receives panels. Both tells fire on the composite exactly as they do today.

**It would also make a rule *more* false.** The geometry finds the active trigger through a private DOM handshake: `root.querySelector('[data-connected-tab][data-state="active"]')` at `:157`, against a stamp at `:371`. Split across components, that has to become published API — slotted rail content must carry those attributes so the pane can find its notch. "Layouts and Surfaces know nothing about their contents" goes from private-to-one-file to false-by-declaration.

The full analysis, including the one clean seam (the compact mobile picker) and why it is a move rather than a fix, is in the proposal I sent Norbert.

## What this PR does

**Deletes two dead props.** `panelClassName` and `style`. Verified by grep across every call site: `panelClassName` appears nowhere in `src` except the component and the story's invented args type, and **zero** callers pass `style`. The membrane drops from seven props to five, three of which are the ordinary tab contract.

**The story stops describing a component that does not exist.** It cast `ConnectedTabs` to an invented args type because `items` is required while the story built items from three knobs (`contentHeight`, `animateDimensions`, `showPanelTitle`) that are not props at all.

Those knobs were never props. They are what the example panels are *made of*. So a helper builds panels and `args` pass them, which is exactly what the sibling `SectionedSurface.stories.tsx` already does. The cast then becomes unnecessary rather than merely unwanted, and the generic types itself by inference from args, the way `PreviewChoice.stories.tsx` already does with the repo's only other generic component.

**No cast, no `render`, no invented type.**

## What is lost, said plainly

The **Content height** range control. A knob that rebuilds panels per keystroke needs a render function, which is what the invented args type existed to serve. The width axis, which is the one that redraws the contour continuously, still works from the viewport toolbar, and `ResizeDrivenGeometry`'s description no longer points at a control that is gone.

If the live height demo is worth keeping, the shape is a per-story `render` over a `.stories.fixture.tsx` component owning the slider. `OverflowPreserved` already has a per-story render and needs no cast, so that route stays open.

## Verified

Every story rendered, not just executed: `ContentDrivenHeight` builds its 720px panels from args and measures a 760px panel with the contour and glass surface intact.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 720 unit tests, 351 storybook tests. The storybook suite runs every play function, including the mobile picker's stepper interaction.

### DIFF
--- a/src/app/ui/surface/ConnectedTabs.stories.tsx
+++ b/src/app/ui/surface/ConnectedTabs.stories.tsx
@@ -2,7 +2,6 @@ import { Box } from '@mantine/core';
 import preview from '@sb/preview';
 import { BookOpen, CircleUserRound, Settings } from 'lucide-react';
 import { useState } from 'react';
-import type { ComponentType, CSSProperties } from 'react';
 import { expect, userEvent, waitFor, within } from 'storybook/test';
 
 import { ConnectedTabs } from './ConnectedTabs';
@@ -12,14 +11,11 @@ import { SurfaceFiller } from './SurfaceFiller.stories.fixture';
 
 type ExampleTab = 'overview' | 'people' | 'settings';
 
-interface ConnectedTabsStoryArgs {
-  value?: ExampleTab;
-  onValueChange?: (value: ExampleTab) => void;
-  items?: ConnectedTabsProps<ExampleTab>['items'];
-  ariaLabel?: string;
-  className?: string;
-  panelClassName?: string;
-  style?: CSSProperties;
+/**
+ * What the example panels are made of.
+ * These are the story's own knobs, not props: the component takes finished panels, so a story that wants taller ones builds taller ones rather than asking the component for them.
+ */
+interface ExamplePanels {
   contentHeight?: number;
   animateDimensions?: boolean;
   showPanelTitle?: boolean;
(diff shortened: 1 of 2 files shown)


ndelangen (commented):
Orchestrator review via an independent agent, post-merge concurrence: no defects. The diff is exactly the recorded no-split decision (membrane 7 to 5, net 44 lines), the two dropped props were verified dead by independent enumeration of all seven call sites including the generic-syntax ones naive grep misses, the story types by inference against the real membrane, and DOM output is byte-identical for every caller. Two story-file convention observations predate the PR and are informational only. All LLM reviewers were quota-blocked; this pass substituted.

Issue #690, opened 2026-08-23
Reset leaves the validation header open with nothing in it


ndelangen:
Found while judging the `WorkbenchLayout` finding on #629. Not a kit-taxonomy problem, so it is not part of the kit wave.

## What happens

Reset an editor page whose draft has warnings, back to a clean baseline, and the validation header stays open with nothing in it: the masthead band keeps showing the "Needs attention" strip and zero chips, until the author happens to blur a field for some unrelated reason.

`ValidationHeader` renders its title strip unconditionally (`ValidationHeader.tsx:53`), so an empty warning list is a visible band, not an empty one.

## Why

`useValidationHeaderOpen(count, settleTick)` is a deliberate latch: warnings open it immediately, an empty list only closes it on a settle signal, so the layout never jumps above the sticky toolbar mid-keystroke. That asymmetry is right.

Reset is not a keystroke, and it emits no settle signal. `onReset: () => setDraft(baseline)` replaces the draft and touches nothing else, in all twelve editor routes.

The blur that comes from clicking the Reset button does not save it, because it lands one render too early: focusout fires while the old warnings still stand, so the settle effect sees a non-zero count and correctly declines to close. The reset lands in the next render, dropping the count to zero with no settle to follow.

## Proof

Driving the hook through the real sequence:

| step | count | settleTick | open |
|---|---|---|---|
| warnings stand | 2 | 0 | true |
| clicking Reset blurs the field | 2 | 1 | true |
| Reset restores the baseline | 0 | 1 | **true** |
| some later unrelated blur | 0 | 2 | false |

## Fix

Reset is a settle: discrete, deliberate, and exactly as much a "the draft has stopped moving" signal as the chapter switch that already counts as one. So the reset handler emits one.

The wiring is duplicated across twelve routes (`grep -rn settleTick src/app/routes`), which is the same twelve-copy shape #641 keeps running into. Worth deciding whether the fix is twelve one-line edits or whether the settle counter and its reset belong together in one hook the routes call.

ndelangen:
## Browser pass: the symptom reproduces on every surface tried, and the diagnosis in this ticket is exactly right

Run against `main` at `becd9255d42`, in a fresh worktree, on a real Chromium at 1440x900. This closes the gap left by the #841 sweep, which traced everything through source and ran nothing.

[...]

PR #691, opened 2026-08-23
The settle handler moves off the layout and onto the editors that own the idiom

ndelangen:
Closes the `WorkbenchLayout` item on [#629](https://github.com/ndelangen/dunezone/issues/629), the last of K4. **Based on `main`, not stacked.** Read the diff with `-w`: the real change is **23 lines added, 12 removed**, and the rest is the reindent from one new level of nesting.

## The #639 half does not bite, and that is worth stating

The finding was parked on ["State the obligation on widget state that mirrors a caller's draft"](https://github.com/ndelangen/dunezone/issues/639) so the `onSettle` threading could be judged against **derive, don't mirror**. It comes back clean, and the rule is the reason why rather than an excuse:

> A **widget may hold no state** whose correctness depends on the history of the draft its caller owns.

The five editors hold **no state at all**. `draft`, `settleTick` and the open/closed latch all live in the route, in the same component, so nothing is mirrored across a membrane. And what the latch tracks is **focus** history, not draft history: blur is not a draft change, and no draft can express "the author left a field."

So this was never a state-mirroring problem. What survives is the taxonomy half.

## What was actually wrong

`WorkbenchLayout.Workbench` took `onBlurCapture` and hung it on its own grid div. The prop's type is honest DOM (`FocusEventHandler<HTMLDivElement>`) and carries no domain knowledge, which is why this is a close call rather than an obvious break.

The tell is the JSDoc. To justify the prop, the layout's own documentation had to say:

> The blur handler rides the grid because settling a draft when focus leaves the form is the editors' shared idiom.

A layout that cannot document itself without naming its callers' domain knows about its contents. That sentence is gone, and `Workbench` is chapters beside a rail again.

## Where it went

Each editor catches focus leaving its own fields. This is not a new pattern: `FactionFormFields.tsx:499` already does exactly this on its own element, and `WorkbenchLayout` was lifted from that editor's stylesheet in the first place. The five now match the one.

Attaching it to the chapters column rather than the whole grid is also the more precise statement of the intent, "focus left the form."

## Verified, because a wrapper element is exactly the kind of change that passes every gate and still moves the page

**Geometry, measured live.** Rendering the token editor and A/B-ing the same DOM, unwrapping the div and re-measuring, at five widths spanning every container step:

| root width | grid columns | match with/without wrapper |
|---|---|---|
| 1100px | 828px / 272px | identical |
| 700px | 524px / 176px | identical |
| 560px | 384px / 176px | identical |
| 430px | 318px / 112px | identical |
| 320px | 208px / 112px | identical |

Every box byte-identical, including where the rail steps down. `ConnectedTabs` declares its own `container: connected-tabs / inline-size`, so the query boundary does not move, and nothing selects `.chapters > *`.

**The event, in a real browser.** A `focusout` from a real field reaches the wrapper **in the capture phase**, which is what `onBlurCapture` binds to.

**Nothing stopped settling.** All five rails hold only `CanvasScale` proofs and `Text`. Counted in the live DOM: **zero** focusable elements. So no blur that settled before fails to settle now.

**One test, pinned.** `onSettle` is documented as firing on field blur and the element carrying it just moved, so `TokenEditor.test.tsx` asserts it. I checked the test is not vacuous: removing the handler fails it. Note the event, since this cost me a false alarm: React implements `onBlur` over `focusout`, so `fireEvent.blur` dispatches a non-bubbling `blur` that never arrives.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 721 unit tests, 351 storybook tests.

## Found on the way, not fixed here

[#690](https://github.com/ndelangen/dunezone/issues/690): Reset leaves the validation header open with nothing in it, in all twelve editor routes. Real, reproducible, and not a kit-taxonomy problem, so it is not in this PR.

### DIFF
--- a/src/app/ui/layout/WorkbenchLayout.tsx
+++ b/src/app/ui/layout/WorkbenchLayout.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx';
 import { Children, isValidElement } from 'react';
-import type { FocusEventHandler, PropsWithChildren, ReactNode } from 'react';
+import type { PropsWithChildren, ReactNode } from 'react';
 
 import styles from './WorkbenchLayout.module.css';
 
@@ -18,12 +18,8 @@ function Rail(_: PropsWithChildren): null {
  * The space at the narrow steps comes from whatever the chapters column holds giving width back, and then from the rail itself becoming a thumbnail.
  * Its desk stretches children to the rail's width, so a proof fills the rail rather than shrinking to content.
  * The desk holds however many artifacts an editor stacks, and the count may change while mounted: a token draws two faces and loses one when its backside becomes a reference.
- * The blur handler rides the grid because settling a draft when focus leaves the form is the editors' shared idiom.
  */
-function Workbench({
-  children,
-  onBlurCapture,
-}: PropsWithChildren<{ onBlurCapture?: FocusEventHandler<HTMLDivElement> }>) {
+function Workbench({ children }: PropsWithChildren) {
   let chapters: ReactNode = null;
   let rail: ReactNode = null;
 
@@ -42,7 +38,7 @@ function Workbench({
 
   return (
     <div className={styles.region}>
-      <div className={styles.workbench} onBlurCapture={onBlurCapture}>
+      <div className={styles.workbench}>
         <div className={styles.chapters}>{chapters}</div>
         <div className={styles.rail}>
           <div className={styles.desk}>{rail}</div>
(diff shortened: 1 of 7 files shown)


ndelangen (commented):
Orchestrator review via an independent agent (all LLM reviewers quota-blocked, so this pass substituted): approved. The layout's membrane is slots-only at head; the blur-capture delta was traced subtree by subtree across all five editors and both route variants, and no focus path exists through the gap (nothing focusable in any Rail proof component; portaled pickers stay covered through React's tree; the toolbar was outside the old grid too). The test pins the contract through a real event path and would pass under either placement, which is the right shape.

One residual risk, non-blocking and documented inline in each widget: a future focusable added to a Rail slot will sit outside the settle subtree that the old grid would have caught. The failure mode is a delayed validation tick, not data loss. If the editors wave restructures these files anyway, folding the rail into the settle boundary then is the cheap moment.

PR #695, opened 2026-08-23
Rulesets adopt About through a compatible production widen

ndelangen:
Tracks #694.

## Summary

- Make `about` the authored Ruleset field while keeping the required legacy `description` column for the compatibility window.
- Normalize reads to About and dual-write both fields so the current and previous Workers can coexist during deployment and rollback.
- Backfill the exact stored value, including Dreamrules' empty string, and add a verifier that blocks retirement while either field is missing or the values differ.
- Rename the create and settings forms to About, keep the 50-character authoring floor, and always render the detail-page section with the standard empty-state copy.

## Migration and rollback

- Deploy order: widened Convex functions and schema, then the backfill and verifier, then the Worker.
- The backfill is idempotent and never overwrites an existing `about` value.
- Rollback order: restore the previous Worker first, then restore the previous Convex release if needed. Both function generations write identical values during the compatibility window.
- The narrow guard belongs to the later retirement ticket, after these migration identities exist in production.

## Verification

- `bun run lint`
- `bun run format:check`
- `bun run typecheck`
- `bun run test` (109 files, 725 tests)
- `bun run check:app-layout`
- `bun run check:convex-skip`
- `bun run check:css-orphans`
- `bun run verify:images`
- `bun run verify:vectors`
- `bun run publisher:release:verify`
- Headless disposable-local `e2e/ruleset-lifecycle.spec.ts` (2 tests)
- Rebased on authoritative `real-origin/main` at `bfe9ade3275c989be5cb74543ebb7787dc62f803`


convex/rulesets.ts
```diff
@@ -30,19 +31,46 @@ async function getRulesetById(ctx: QueryCtx | MutationCtx, id: Id<'rulesets'>) {
...
+
+/**
+ * Accepts the new About argument and the old Worker argument during the compatibility window.
+ * Supplying both is allowed only when they name the same trimmed prose, so no caller can make the dual-written columns disagree.
+ */
+function parseCompatibleRulesetInput(input: CompatibleRulesetInput) {
+  if (input.about === undefined && input.description === undefined) {
+    throw new Error('Ruleset About is required');
+  }
+  if (input.about !== undefined && input.description !== undefined && input.about.trim() !== input.description.trim()) {
```
ndelangen: Fixed in fdacb62c. The parser now collects the one or two supplied values and compares their trimmed Set size, leaving only simple empty and mismatch branches. The focused compatibility and migration suites still pass.

Issue #698, opened 2026-08-24
Work the data wave


ndelangen:
Execute the data wave, first of the three post-kit waves per the ordering decision on #697. The ranked list from the subscriptions audit (#632), worked top down:

1. The catalogue faction-row narrowing: decided and measured on #642 (124 KB to 28 KB on /factions; server projection and CanonicalFactionClientSchema narrow together in one PR; browser pass on the three consuming surfaces; the measurement comment is the spec).
2. The shell's per-page memberships join: profiles.current computes default_group_options for every _app page to feed one Select on profile edit; also collapse useCurrentProfile's two queries where one serves.
3. The group-detail eager assign popover: the one picker-contract breach; convert to the defer-to-open inline Picker shape AGENTS.md specifies.
4. The fat detail bundles: groups.detailBySlug shipping full faction blobs for a link list, the FAQ create page holding a whole ruleset bundle for a name and an id, rulesets.detailPageBySlug on the edit route.
5. The assets landing's unrendered owner field.

Charter constraints: no schema or index changes without individual sign-off; one session does database-touching work at a time, coordinated through the orchestrator; every PR small, reviewed by the orchestrator, merged by Norbert; visual surfaces get browser passes with real data.

ndelangen:
Item 3 decided (Norbert): shape 2, the shell-and-picker split. AssignPopover reduces to the kit shell (trigger and pane, mounting content only while open) and a new GroupAssignPicker in src/app/pickers owns the query, fetching on mount, lazy because the shell gates mounting. This is the FactionPicker-inside-FactionLoadPopover precedent applied, per AGENTS.md's own text (a Picker renders through the kit, an AssignPopover named as the example). Rejected: a render-when-open slot on the existing component, which would grow a kit Control's membrane with a content slot and smudge the receiver-versus-producer line. The earlier judge objection targeted a kit-into-two-kit-components split, which this is not.

ndelangen:
Resolved: the ranked five are delivered, plus two defect fixes found in flight. Item 1 the catalogue narrowing (#699, 76 percent off the catalogue payload), item 2 the shell memberships join (#702, with the loading-window follow-up #703), item 3 the assign shell-and-picker split (#713, with the arrow fix #714 and the focus-trap pane-shape fix riding), items 4 and 5 folded where the audit routed them (the fat-bundle and owner-field reductions landed inside the narrowing and session work; anything residual is measured as gone in the #632 re-checks on the PRs).

[...]

PR #699, opened 2026-08-24
The catalogue ships what it draws, not the whole authored faction

ndelangen:
First item of [#698](https://github.com/ndelangen/dunezone/issues/698), decided and measured on [#642](https://github.com/ndelangen/dunezone/issues/642). **Based on `main`, not stacked.** No schema or index change.

Catalogue-shaped faction rows carried the whole authored blob to three surfaces that render six of its thirteen `data` fields.

## One mask, four derivations

The narrowing exists in exactly one place, a field mask in `src/shared/factions/schema.ts`. Everything else follows from it:

| | derived from |
|---|---|
| `CatalogueFactionStoredSchema` | the mask, off the canonical stored schema |
| `CatalogueFactionClientSchema` | the mask, off the canonical client schema (stays loose) |
| `catalogueFactionDataValidator` | `zodToConvex` of the stored variant, the existing ADR-0002 route |
| `CatalogueFaction` (server) | `Infer` of the wire validator |
| `FactionCatalogueRow` (client) | `FunctionReturnType` of `factions.cataloguePage` |

So there is no second list of field names to fall out of step, and the client type cannot claim a field the server stopped sending.

Both build sites, the catalogue loader and the ruleset faction rail, were spreading the row independently. They now go through one `toCatalogueFaction`.

`factions.getBySlug` is untouched: detail pages and the editor keep the full blob through their own contract.

## The stale-tab consequence, stated plainly

Narrowing **is** breaking for a tab running the previous bundle. The client boundary's tolerance is one-directional: `CanonicalFactionClientSchema` is a `looseObject`, so it accepts unknown *extra* fields, but it still requires the ones it declares. I verified this rather than assuming it, and the reverse case fails as expected.

That is already handled by design here. `parseClientBoundary` throws `StaleClientDataError`, documented in that file as "the stale-tab signal", and the router renders it as a refresh prompt instead of a crash. **This is the case that mechanism exists for**, so no two-phase widen-then-narrow deploy is needed.

## Verified

**The wire shape matches the spec exactly**, checked against the validator itself rather than by reading:

- row: `_id, created_at, data, rulesets, slug, updated_at`
- `data` kept: `background, complexity, hero, leaders, logo, name`
- `data` dropped: `colors, decals, extras, planet, rules, themeColor, troops`

Those seven are precisely the ones the measurement identified as dead on all three surfaces.

**The saving, re-measured independently on today's live data** rather than quoted from the ticket. Real `factions.cataloguePage` payload from the dev deployment, projected to the narrowed shape:

| | |
|---|---|
| rows | 35, plus 2 spotlight rows shipped twice |
| full payload | 130.1 KB |
| narrowed payload | 31.0 KB |
| saving | **99.1 KB (76.2%)** |

`rules` alone is 75.1% of blob weight. This lands where #642 predicted (124 KB to 28 KB); the small drift is data added since. The saving recurs on every subscription resend, not just first load.

**Browser pass on all three consuming surfaces**, with the narrowed functions actually deployed:

| surface | result |
|---|---|
| `/factions` | 37 faction cards plus 2 spotlights, no stale-data prompt |
| `/rulesets/dreamrules` | all 11 rail factions render |
| `/profiles/central` | faction list renders |

All six kept fields draw. Bene Gesserit shows hero "Gaius Helen Mohiam" at strength 5, three leaders, the name, background artwork, the complexity glyph, and its "Dreamrules" caption. `FactionCard` destructures exactly the six kept fields and reads nothing else.

**The narrowing is pinned, and the pin is not vacuous.** `factions.catalogue.test.ts` asserts the row and `data` key sets, and asserts the stored fixture really does carry the dropped fields so it cannot pass for the wrong reason. Re-widening the projection fails it, and the Convex return validator rejects the extra fields too.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 727 tests.

**Shared dev is back on `main`.** The browser pass required deploying these functions to `dev:tame-raccoon-541`. I have since pushed `main`'s functions back and confirmed by querying the deployment that it serves the full blob again, rather than assuming the push landed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved faction catalogue data handling by returning only the fields needed for catalogue views.
  * Standardized faction data across catalogue, ruleset, and profile pages.
  * Added validation to ensure catalogue entries contain the expected information.
  * Preserved additional faction details in stored data while excluding them from catalogue responses.

### DIFF
--- a/convex/factions.catalogue.test.ts
+++ b/convex/factions.catalogue.test.ts
@@ -94,6 +94,37 @@ describe('faction catalogue page', () => {
     expect(catalogue.spotlights.freshlyUpdated?.data.name).toBe('Recently updated');
   });
 
+  test('ships what the catalogue draws and leaves the rest of the blob behind', async () => {
+    const t = convexTest(schema, modules);
+    await t.run(async (ctx) => {
+      const ownerId = await ctx.db.insert('users', { name: 'Narrowing owner' });
+      await ctx.db.insert('factions', {
+        owner_id: ownerId,
+        data: { ...assetPublishingFaction, name: 'Full blob' },
+        slug: 'full-blob',
+        created_at: '2026-07-20T00:00:00.000Z',
+        updated_at: '2026-07-20T00:00:00.000Z',
+        is_deleted: false,
+        group_id: null,
+      });
+    });
+
+    const catalogue = await t.query(api.factions.cataloguePage, {});
+    const row = catalogue.factions[0];
+
+    expect(Object.keys(row ?? {}).sort()).toEqual(['_id', 'created_at', 'data', 'rulesets', 'slug', 'updated_at']);
+    expect(Object.keys(row?.data ?? {}).sort()).toEqual([
+      'background',
+      'complexity',
+      'hero',
+      'leaders',
+      'logo',
+      'name',
+    ]);
+    /* The stored fixture really does carry the dropped fields, or the assertions above pass for the wrong reason. */
+    expect(Object.keys(assetPublishingFaction)).toEqual(expect.arrayContaining(['rules', 'troops', 'colors']));
+  });
+
   test('rejects malformed faction data at the query boundary', async () => {
     const t = convexTest(schema, modules);
     await t.run(async (ctx) => {
--- a/convex/factions.ts
+++ b/convex/factions.ts
@@ -14,7 +14,7 @@ import {
 import {
   factionDetailPageValidator,
   factionValidator,
-  factionWithRulesetsValidator,
+  catalogueFactionValidator,
   rulesetSummaryValidator,
 } from './lib/collaborativeAccessValidators';
 import { resolveGroupAssignmentForCreation } from './lib/defaultGroupPreference';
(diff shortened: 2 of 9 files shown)


ndelangen (commented):
Orchestrator review, commits enumerated (one): approved pending CI. Verified in the diff: the single catalogueFactionMask with all three schemas derived by pick, the wire validator swapped at every catalogue-shaped return (cataloguePage rows and both spotlights, the ruleset rail, the profile list), the row narrowed to the spec's exact field union, and the non-vacuous pin. The browser pass covered all three consuming surfaces against the actually-deployed narrowed functions, the live re-measurement (130.1 to 31.0 KB, 76.2 percent) confirms the ticket's numbers on today's data, and the stale-tab section documents the one-directional tolerance and the StaleClientDataError refresh path where the merge reviewer will read it. The restore-from-verified-equivalence push, declining to use a checkout the pusher could not inspect, is noted with approval as the standard for shared-infrastructure operations.

ndelangen (commented):
Orchestrator re-review at f30da92bb4c (approval re-anchored; the earlier approval named no SHA and the diff moved under it). Approved. The exactness fix is verified: the type derives from the strict variant while the parser stays loose, matching toFactionEntry's existing split, and the @ts-expect-error assertion pins the derivation side so the hole cannot quietly reopen. The manifest regeneration commit is the expected identity consequence of touching the shared schema module. My original review endorsed the mask design without catching that .pick() carries a looseObject's catchall into the inferred type; the independent adversarial pass caught what I did not, and finding 2's correction of the PR body's own stale-tab claim (accurate on one of three surfaces, filed as #700) is the honest-record standard. From here my reviews name the SHA they reviewed, and an approval lapses when the head moves.

PR #702, opened 2026-08-24
The shell stops paying a memberships join for a Select it never renders

ndelangen:
Item 2 of [#698](https://github.com/ndelangen/dunezone/issues/698), from finding 1 of the subscriptions audit [#632](https://github.com/ndelangen/dunezone/issues/632). **Based on `main`, not stacked.** No schema or index change.

Every `_app` page holds the session read through `SiteNavigation`, and that read computed `default_group_options`: an active-memberships query plus a group read per membership. One Select, on one tab of the profile edit form, consumes it.

## Two changes

**The preference moved to the page that renders it.** New `profiles.defaultGroupPreference`, held by `EditableProfilePage` rather than by `ProfileSettingsPage`. That placement is deliberate: the outer component also renders for signed-out visitors and for people viewing someone else's settings, and neither should subscribe to a memberships join.

**The session pair collapsed.** `profiles.currentUserId` and `profiles.current` were two subscriptions answering one question. They are now `profiles.session`, returning `{ userId, profile }`. Both queries also gained return validators, which neither had.

## Three things worth reviewing rather than skimming

**The envelope is never `null`, on purpose.** A signed-out viewer and a signed-in viewer with no profile row yet have to stay distinguishable, or `useCurrentProfile`'s bootstrap never fires and a new OAuth account never gets a profile. `{ userId: null, profile: null }` and `{ userId, profile: null }` are different states and the shape preserves that.

The client mirror of the same trap: `toLiveQueryResult` derives `isPending` from `liveData === undefined`, so writing `session?.profile` would make a signed-out viewer read as permanently pending. The explicit ternary keeps today's behaviour exactly.

**The reconcile effect gates on loaded, not on non-empty.** The tempting `(options ?? []).some(...)` reads the loading window as "you are in no Groups", clears a valid saved default, and does not set `defaultGroupChanged` — so `isDirty` stays false and nothing on screen says it happened. It gates on `!defaultGroupOptions` instead.

**`session.profile.default_group_id` is now the raw column**, not the sanitized projection, so it can point at a Group the viewer has left. Both write paths (`updateCurrent` and `resolveDefaultGroupForCreation`) already re-check independently, and the form corrects a stale display once options arrive. Flagged because it is a real behaviour change, not because it is unsafe.

## Considered and rejected: the Picker shape

`ConnectedTabs` unmounts inactive panels, so a `DefaultGroupPicker` inside the `defaults` panel would be strictly lazier than a route-held query. It is rejected because the reconcile effect needs the option list *while that panel is unmounted*, and feeding it upward would make a Picker read page state, which is the line AGENTS.md draws. The audit's finding 1 specifies a route-held query; the defer-to-open shape is finding 2, a different item.

## Verified

**The server contract, against the deployment** rather than from the source:

| query | anonymous response |
|---|---|
| `profiles:session` | `{ userId: null, profile: null }` |
| `profiles:defaultGroupPreference` | `{ default_group_id: null, default_group_options: [] }` |
| `profiles:current` | gone — "Could not find public function" |
| `profiles:currentUserId` | gone — same |

The app ran correctly against exactly those functions, which is also the proof the client is not still asking for the old pair: it could not be, they do not exist.

**In the browser, with the change deployed:**

- Visiting `/profiles/central/edit` as a non-owner subscribed to **nothing at all**, and `defaultGroupPreference` in particular was never requested. Recorded off the Convex websocket, not inferred.
- The shell settled to "Login" rather than a pending state, which is the signed-out half of the never-null envelope working end to end.
- `/factions` rendered 39 cards, no console errors.

**The split is pinned and the pin is not vacuous.** A test asserts `session.profile` no longer carries `default_group_options`, using a fixture whose member really is in a Group. Putting the projection back on `session` fails it.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:convex-skip, check:css-orphans, check:app-layout, 727 tests.

## Not verified, and I would rather say so

The signed-in owner path was not exercised in a browser: this deployment's local origin has no session, and I am not going to handle credentials to get one. That path is covered by the unit test (mocked hook, clicks through to a seeded Group) and by `convex/defaultGroupPreference.test.ts`, which exercises the new query with a real member in a real Group.

I also considered adding an e2e assertion that opens the combobox, and did not: the e2e user has no Group seeded, so no assertion there can tell "options loaded, user has none" apart from "options never loaded". It would look like coverage without being any.

## What this does not fix

The per-membership fan-out is unchanged — one `db.get` per active membership, under the existing `take(500)` cap. This is a subscription-scope change, not a read-count change: the join now runs on one page instead of every page.

## Summary by CodeRabbit

* **Improvements**
  * Improved profile loading to clearly distinguish signed-out users, authenticated users without a profile, and active profiles.
  * Profile setup now proceeds automatically for authenticated users who do not yet have a profile.
  * Default Group settings now use current membership information, preventing unavailable Groups from being selected.
  * Profile editing provides refreshed Group options and safely handles loading or outdated selections.

* **Documentation**
  * Updated authentication guidance for the revised profile loading and default Group preference experience.

### DIFF
--- a/convex/accountLifecycle.widen.test.ts
+++ b/convex/accountLifecycle.widen.test.ts
@@ -83,7 +83,7 @@ describe('account lifecycle widen rollout', () => {
     });
 
     const pending = t.withIdentity({ subject: ids.userId });
-    await expect(pending.query(api.profiles.currentUserId, {})).resolves.toBeNull();
+    await expect(pending.query(api.profiles.session, {})).resolves.toEqual({ userId: null, profile: null });
     await expect(pending.mutation(api.profiles.bootstrapCurrent, {})).rejects.toThrow(/Not authenticated/);
     await expect(t.query(api.profiles.getBySlug, { slug: 'pending-user' })).rejects.toThrow(/not found/);
     await expect(t.query(api.profiles.newestDiscoverable, {})).resolves.toEqual([]);
--- a/convex/defaultGroupPreference.test.ts
+++ b/convex/defaultGroupPreference.test.ts
@@ -94,10 +94,15 @@ describe('default Group preference', () => {
 
     const legacyProfile = await t.run(async (ctx) => await ctx.db.get('profiles', ids.memberProfileId));
     expect(legacyProfile).not.toHaveProperty('default_group_id');
-    await expect(member.query(api.profiles.current, {})).resolves.toMatchObject({
+    await expect(member.query(api.profiles.defaultGroupPreference, {})).resolves.toMatchObject({
       default_group_id: null,
       default_group_options: [{ id: ids.groupId, name: 'Default Group' }],
     });
+
+    /* The point of the split (#698 item 2): every `_app` page holds `session`, so the memberships join
+       must not ride along on it. This member is in a Group, so a regression here would carry a populated list. */
+    const session = await member.query(api.profiles.session, {});
+    expect(session.profile).not.toHaveProperty('default_group_options');
 
     await expect(
       member.mutation(api.profiles.updateCurrent, {
@@ -117,7 +122,7 @@ describe('default Group preference', () => {
       profile: { default_group_id: ids.groupId },
       default_group_unavailable: false,
     });
-    await expect(member.query(api.profiles.current, {})).resolves.toMatchObject({
+    await expect(member.query(api.profiles.defaultGroupPreference, {})).resolves.toMatchObject({
       default_group_id: ids.groupId,
       default_group_options: [{ id: ids.groupId, name: 'Default Group' }],
     });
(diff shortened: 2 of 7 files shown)


ndelangen (commented):
Orchestrator review at bcc70dd1324, branch declared at rest. Approved pending CI. Verified in the diff: the never-null session envelope with its reasoning in the JSDoc and both branches distinguishable; the reconcile effect gating on options-not-yet-loaded rather than empty-list, so the loading window cannot clear a saved default; the preference query held by EditableProfilePage with the non-subscription proven off the websocket for a non-owner; the raw-column display correction understood as display-only with both write paths re-checking server-side. Checked the house reset rule: this editor has no reset affordance and the touched states predate the PR, so nothing bites.

Both judgment calls endorsed. Declining the e2e assertion is the over-testing preference applied correctly: the seeded e2e user cannot distinguish options-loaded-and-empty from options-never-loaded, so the assertion would read as coverage without being any, and the Convex test against a real member is the honest layer. Declining to handle credentials for a signed-in browser pass is right; the consequence is stated plainly in the body rather than left to be noticed: the signed-in owner's Select path has run under tests but not in a browser, and it deserves the maintainer's one manual glance after this deploys.

PR #705, opened 2026-08-24
CanvasScale fits its canvas in CSS and stops measuring

ndelangen:
The last item of [#641](https://github.com/ndelangen/dunezone/issues/641), which this closes. **Based on `main`, not stacked.** Net 22 lines removed.

Norbert's decision is [recorded on the ticket](https://github.com/ndelangen/dunezone/issues/641#issuecomment-5394223591): option C, simplify now, no exemption sentence and no scheduled revisit, because current and future simplicity outweighs a temporary niche graceful failure and a revisit is itself a cost.

## What changes

`CanvasScale` measured its container with a `ResizeObserver` and wrote `transform: scale(width / canvasWidth)` in JS. The frame is now its own container, so `100cqw` **is** its width, and the division happens in CSS:

```css
transform: scale(calc(100cqw / var(--canvas-width)));
```

The observer goes, along with two pieces of state, an effect, a callback ref and the `width > 0` render gate.

Two consequences worth naming:

- The canvas is taken **out of flow**, so the frame's height comes from its `aspect-ratio` alone and never from the unscaled child it clips. Without that the frame would be sized by a 1263px-tall child.
- The proof is present **on first paint** rather than after a measurement lands, which removes a frame of empty rail.

## Accepted Firefox tail

Firefox 154 and older lack `calc()` typed arithmetic and render the proofs as a **silently plausible close-up crop** rather than anything obviously broken. Firefox 155 ships the feature on 1 September 2026.

That is the accepted consequence of the decision, not an open question. Measurements across all three engines, the isolation of which feature is missing, the Bugzilla and release-note links, and screenshots of the failure are all in [the evidence comment](https://github.com/ndelangen/dunezone/issues/641#issuecomment-5394223591); this PR does not re-argue them.

## Three test files change, and the reason is the interesting part

`BundleEditor`, `TokenEditor` and `DeckEditor` each carried a `ResizeObserver` stub whose comment said it existed because an unmeasured `CanvasScale` "renders no children at all", so a dead stub would let assertions about a rail proof pass against a DOM that never contained one. That hazard is now gone.

The stub is not, and I checked rather than assumed: removing it entirely fails the suite with `ReferenceError: ResizeObserver is not defined`, because `PreviewChoice` constructs one without guarding for its absence. But making it **inert** passes, so it no longer needs to fire. All three now carry a two-line stub and a comment that says what is actually true.

## Verified

**In Chromium, at four widths**, driving the real story and measuring:

| container | frame | canvas on screen | computed transform |
|---|---|---|---|
| 450px | 450x632 | 450x632 | `matrix(0.5, …)` |
| 300px | 300x421 | 300x421 | `matrix(0.333333, …)` |
| 180px | 180x253 | 180x253 | `matrix(0.2, …)` |
| 90px | 90x126 | 90x126 | `matrix(0.1, …)` |

Every scale is exactly container ÷ 900, the aspect holds, and the canvas fills the frame at each size. The scaling is now continuous rather than quantised by observer callbacks.

**In situ on a rail**, the token editor story draws both proofs in 256px frames at `matrix(0.284444, …)`, which is 256 ÷ 900, with their artwork intact.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:app-layout, 727 unit tests, 351 storybook tests. The renderer manifest is unchanged; `CanvasScale` is not in the capture bundle's import graph.

## Summary by CodeRabbit

* **Improvements**
  * Canvas previews now scale smoothly within their available frame while preserving aspect ratio, clipping, rounded corners, and pointer interactions.
  * Canvas frames render immediately for a more consistent visual experience.

* **Tests**
  * Updated editor test coverage to use stable resize behavior, reducing layout-dependent results in test environments.

### DIFF
--- a/src/app/ui/layout/CanvasScale.tsx
+++ b/src/app/ui/layout/CanvasScale.tsx
@@ -1,10 +1,13 @@
-import { useEffect, useState } from 'react';
-import type { PropsWithChildren } from 'react';
+import type { CSSProperties, PropsWithChildren } from 'react';
 
 /**
  * Fits a fixed-size canvas, a game renderer drawing in card-space pixels, to the container's width: the frame keeps the canvas aspect and clips it while the canvas renders at native size and scales to fit, down in a rail and up in a wide mount, which stays crisp because the renderers are DOM and vector rather than raster.
+ * The fit is CSS, not measurement: the frame is its own container, so `100cqw` is its width, and dividing that by the canvas width gives the scale directly.
+ * Nothing here re-renders on resize, and the canvas is present on first paint rather than after a measurement lands.
+ * The canvas is taken out of flow so the frame's height comes from its aspect ratio alone, never from the unscaled child it clips.
  * Pure placement;
  * the caller decorates the frame (shadow, anything bespoke) through `frameClassName`, and `rounded` is the one shared decoration: the corner treatment every rail proof wears (Norbert, 2026-08-21), kept here so five rails cannot drift apart.
+ * The browser floor this technique sets is recorded on «Work the kit compliance wave» (#641).
  */
 export function CanvasScale({
   canvasWidth,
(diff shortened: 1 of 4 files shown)


ndelangen (commented):
Orchestrator review at a32c182f, branch at rest. Approved pending CI. The rewrite is the decision executed exactly: no hooks, no observer, the frame as its own container with scale(calc(100cqw / var(--canvas-width))), the out-of-flow reason stated in the JSDoc, one line pointing at #641, no Firefox history. Verified pointer-events parity with main, and the test-stub changes are the honest consequence: the width-gated render that made a dead stub dangerous is gone, children mount unconditionally in jsdom, and the surviving two-line stub names the one real reason it exists. The measured continuous scaling (container/900 exact at four widths, in-situ matrix 256/900) is the right evidence. On merge this closes the kit compliance wave; the two shared memories describing the old stub hazard will be retired with it.

Issue #707, opened 2026-08-24
Effect and observer census: 31 of 49 sites have a declarative equivalent


ndelangen:
Norbert's position: almost every `useEffect` and `ResizeObserver` is a code smell whose use has to be defended. Filing the evidence, because it came back with a crisper rule than the one we started with, and with a work list.

Every site in `src` outside tests was classified as **replaceable** (a declarative equivalent exists today, named), **earned** (no declarative form, with the reason), or **unclear**. The classifications were then spot-checked against the code, and several were downgraded.

## The numbers

49 sites. **31 replaceable (63%), 17 earned (35%), 1 unclear.**

## The ResizeObserver rule holds exactly, and it is cheap to enforce

8 observers remain after [#705](https://github.com/ndelangen/dunezone/pull/705): **5 replaceable, 3 earned.** The split is not noisy:

- Every observer measuring a box **to compute a style** is replaceable. 5 of 5, no exceptions.
- Every observer measuring **to produce a count or a boolean** is earned. 3 of 3.

So the reviewer's test is one line: **does the number become a style, or a decision?** A style is CSS's job. A decision may be JavaScript's.

## The useEffect rule holds, for a different reason than the one that motivated it

The shape that prompted this, measure-to-style, is not the dominant smell:

| shape | count |
|---|---|
| prop or derived value synced into state | 9 |
| measure-to-style (the CanvasScale case) | 6 |
| document attribute written imperatively that React could render | 4 |
| write that belongs in an event handler | 3 |
| child-to-parent state relay | 3 |
| platform grew a declarative answer since | 2 |
| dead code | 2 |

And where the rule is weakest: **effect-as-subscription and effect-as-cleanup came back 17 for 17 earned.** Not one is a disguised derivation. So the honest formulation is not "effects are suspect" but "**an effect that writes React state is suspect; an effect that subscribes is doing its job**".

The kit already knows this. `src/app/ui/control/useHoldToConfirm.ts:29-35` does the render-phase reset with a comment calling it "the search box's pattern", then reaches for an effect only at `:45` to release an interval.

## What earned looks like here

Worth having as a positive rule, since a prohibition people route around is worse than a test they can apply:

1. **A measurement producing a count that picks a branch.** `SiteNavigation.tsx:165`: how many links fit decides which render inline, which move to the popover, and whether the More button exists.
[...]

PR #708, opened 2026-08-24
The shield preview stops measuring itself

ndelangen:
First item of [#707](https://github.com/ndelangen/dunezone/issues/707)'s declarative wave. Net 25 lines removed.

**Now based on `main`.** It was opened stacked on [#705](https://github.com/ndelangen/dunezone/pull/705) and retargeted once that merged; the diff is the shield change alone.

## Why it was stacked, kept as history

The replacement is "use `CanvasScale`", and before #705 that component rendered **nothing** until its `ResizeObserver` reported a width. `ScaledFactionShield` seeded `frameWidth` at 352 precisely so the shield was painted on first render. Landing this against the old `CanvasScale` would therefore have traded one observer for another **and** introduced a blank first paint. That is why it waited rather than raced, and why anyone tempted to reorder this pair should not.

## What it deletes

`ScaledFactionShield` was #705's own precedent, one commit earlier: a fixed 1608x805 canvas, `transform-origin: top left`, an inline `height: shieldSize.height * scale`, and a `useLayoutEffect` + `ResizeObserver` to keep `scale = frameWidth / 1608` current.

Both halves are already CSS:

- `scale = frameWidth / 1608` **is** `calc(100cqw / 1608px)`
- `height = 805 * scale` **is** `aspect-ratio: 1608 / 805`

Nothing branched on the number. It was only ever a style, which is the test [#707](https://github.com/ndelangen/dunezone/issues/707) settled on.

Gone: the layout effect, the observer, the `frameWidth` state, the 352px seed, and the `useLayoutEffect` import.

## The duplicate that came with it

`.shieldCanvas` hard-coded `width: 1608px; height: 805px`. That is a second, hand-written copy of `shield` in [`src/game/data/sizes.ts`](src/game/data/sizes.ts), which the component **already imports** as `shieldSize`. The class is deleted rather than moved, and the two numbers now have one source.

## The trap, handled

`CanvasScale` writes `width: 100%` as an inline style, which beats a class. So `.shieldFrame` could not keep owning the width.

A `.shieldMount` wrapper takes the width (`min(100%, 22rem)`, and `min(30%, 17rem)` under the `faction-review-plane` container query) plus `flex: 0 0 auto`, and carries the `role="img"` and `aria-label` so the picture is still named. `.shieldFrame` keeps only its `drop-shadow`; `position: relative` and `overflow: hidden` come from `CanvasScale` itself.

## Verified in the browser, both width paths

Driving `FactionSheetReview`'s desktop story and measuring:

| state | mount | frame | scale applied | expected |
|---|---|---|---|---|
| base width | 352px | 352x176 | `0.218905` | 352 / 1608 |
| container query fired | 272px | 272x136 | `0.169154` | 272 / 1608 |

The aspect holds in both, the drop-shadow is still on the frame, and the shield draws its 7 SVGs. The second row needed the review plane forced to 1100x700, since the container query wants 56rem x 34rem and the story's plane is smaller than that by default.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 727 unit tests, 351 storybook tests.

### DIFF
--- a/src/app/widgets/faction-editor/FactionSheetReview.module.css
+++ b/src/app/widgets/faction-editor/FactionSheetReview.module.css
@@ -195,22 +195,15 @@
   padding-bottom: 3.5rem;
 }
 
-.shieldFrame {
-  position: relative;
+/* Owns the width, because `CanvasScale` sets its own to 100% inline and an inline style beats a class. */
+.shieldMount {
   flex: 0 0 auto;
   width: min(100%, 22rem);
-  overflow: hidden;
-  filter: drop-shadow(0 0.75rem 1.2rem rgb(0 0 0 / 32%));
 }
 
-.shieldCanvas {
-  position: absolute;
-  top: 0;
-  left: 0;
-  width: 1608px;
-  height: 805px;
-  transform-origin: top left;
-  pointer-events: none;
+/* Only the decoration: the frame's position and clipping belong to `CanvasScale`. */
+.shieldFrame {
+  filter: drop-shadow(0 0.75rem 1.2rem rgb(0 0 0 / 32%));
 }
 
 .sheetSpread {
@@ -260,7 +253,7 @@
     padding-bottom: 0;
   }
 
-  .shieldFrame {
+  .shieldMount {
     width: min(30%, 17rem);
   }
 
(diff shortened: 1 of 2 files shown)


ndelangen (commented):
Orchestrator review at 6a8964b28b5, base main, branch at rest. Approved pending the settling checks. The census item is executed exactly: the observer, the layout effect, the 352px first-paint seed, the inline height and the hard-coded 1608x805 stylesheet duplicate are all gone; the shield's size is read from @game/data/sizes and written down nowhere else; the flagged inline-width trap is handled by the mount wrapper owning width and role=img, with the reason stated in the stylesheet comment; decoration rides frameClassName as CanvasScale's membrane sanctions. Net minus 25 lines and one less thing that measures. The retarget-then-speak sequence and the ancestry verification before touching the base are noted as the standing rule followed to the letter.

PR #709, opened 2026-08-24
Three drafts re-seed during render instead of through an effect

ndelangen:
Item 7 of [#707](https://github.com/ndelangen/dunezone/issues/707)'s declarative wave. **Based on `main`, not stacked.**

Three controls keep a draft the reader edits, and re-seed it when the committed value changes underneath: `MemberCountInput`, the catalogue's search box, and the complexity range slider. All three did the re-seed in an effect, which means a commit and a paint showing the **stale** draft before the correction lands.

## The repo already had the fix, and a name for it

`useHoldToConfirm.ts:28` calls it "the search box's pattern": compare during render, and React re-runs the component before painting.

```tsx
const [seeded, setSeeded] = useState(value);
if (value !== seeded) {
  setSeeded(value);
  setTyped(value);
}
```

All three now do that. Two `useEffect` imports become unused and go.

## One that is not mechanical

The complexity range compares the **raw prop**, not the parsed range:

```tsx
if (value !== seeded) {
  setSeeded(value);
  setDraft(parseComplexityRange(value));
}
```

`parseComplexityRange` allocates a new object every call, so comparing parsed values would be true on every render and reset the draft forever. The effect version was safe from this only because its dependency array held the raw `value`.

## The branch had no test, so it has one now

`MemberCountInput.test.tsx` gains a case where the committed count changes while an untouched edit is in the field. Both existing tests drive the field and never change `value` from outside, so the re-seed, the exact behaviour this PR rewrites, was uncovered.

I checked it is not vacuous: removing the render-phase reset fails it.

The other two are covered indirectly. The catalogue's re-seed rides the URL and is exercised by `-catalogue.test.ts`; the range slider is a story-only control.

## Proof, and what it does not prove

`MemberCountInput` is the one of the three with stories; the catalogue search box and the complexity range are route-level. All four of its stories, after the change:

| default | at minimum | at maximum | disabled |
|---|---|---|---|
| default | at minimum | at maximum | disabled |

And the draft holding an edit, which is the state a re-seed has to override:

mid-edit, field showing 42

**Being straight about these:** they show the control still renders correctly, which is the parity half of the claim. They cannot show the fix. What this PR removes is a single paint where the field showed the stale draft before the effect corrected it, and a still image cannot show a frame that no longer happens. That half is carried by the test, which fails when the render-phase reset is removed.

The other two are unscreenshotted: the search box's re-seed rides the URL and the range slider is route-level. Their change is the same three lines and the same reasoning.

## Verified

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 729 unit tests, 351 storybook tests.

## A note on the census this came from

While working the items around this one I checked [#707](https://github.com/ndelangen/dunezone/issues/707)'s other claims and two of them do not hold. Details are on that ticket rather than here, but in short: the preview-sheet effects are duplicated, not racing, and the effect itself is earned; and of the icons route's four effects, one is a documented fix for a real bug rather than a React 17 guard. The census is directionally right and its per-item reasoning still needs checking, which is what it asked for.

### DIFF
--- a/src/app/routes/_app/factions/-catalogue.ts
+++ b/src/app/routes/_app/factions/-catalogue.ts
@@ -149,8 +149,13 @@ export function useFactionCatalogueSession(data: Pick<FactionCataloguePageData, 
   const search = parseFactionCatalogueSearch(location.search ?? {});
   const rawSearch = location.searchStr;
   const [draftQuery, setDraftQuery] = useState(search.q ?? '');
-
-  useEffect(() => setDraftQuery(search.q ?? ''), [search.q]);
+  const [seededQuery, setSeededQuery] = useState(search.q ?? '');
+  /* Reset during render, the search box's pattern: the URL is the committed value, and a back button
+     that restores an older query must not paint the previous one first. */
+  if ((search.q ?? '') !== seededQuery) {
+    setSeededQuery(search.q ?? '');
+    setDraftQuery(search.q ?? '');
+  }
 
   useEffect(() => {
     if (!data) {
--- a/src/app/routes/_app/factions/index.tsx
+++ b/src/app/routes/_app/factions/index.tsx
@@ -27,7 +27,7 @@ import { FactionList } from '@ui/list/FactionList';
 import { Surface } from '@ui/surface';
 import { Toolbar } from '@ui/surface/Toolbar';
 import { ArrowDownAZ, ChevronsDown, Filter, Plus, Search, SlidersHorizontal } from 'lucide-react';
-import { useEffect, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
 import type { KeyboardEvent } from 'react';
 
 import { loadFactionCataloguePage, useFactionCataloguePage } from '@db/factions';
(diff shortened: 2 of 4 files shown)


ndelangen (commented):
Orchestrator review at 9f795a76, base main, branch at rest. Approved pending CI. The render-phase reset is the idiom the kit already names, the complexity-range subtlety is handled the only correct way with the constraint written in place (comparing the raw prop because parsing allocates per call; the effect version was accidentally safe through its dependency array, which is exactly the kind of safety that evaporates in refactors), and the new test drives the previously uncovered re-seed branch and was checked to fail without the reset.

The two census corrections on #707 are endorsed as the more valuable half of this hand-off: item 6 reclassified from race-to-dedup (the effect is earned by the census's own writes-outside-React test), item 5 halved with one site defensible and one load-bearing, and the classifier bias named (readier to call replaceable than the code supports) and applied forward to the unstarted items. A census that corrects its own numbers downward mid-execution is worth more than one that stays impressive.

PR #713, opened 2026-08-24
The assign pane stops holding the choices, so a picker can fetch them lazily

ndelangen:
Closes item 3 of [#698](https://github.com/ndelangen/dunezone/issues/698), [decided by Norbert](https://github.com/ndelangen/dunezone/issues/698#issuecomment-5395144442) as shape 2, the shell-and-picker split. **Based on `main`, not stacked.**

## What was wrong

`factions.listOwnedForGroupAssign` and `rulesets.listOwnedForGroupAssign` subscribed when the group page mounted, for every active member, to fill a popover most visits never open.

## The split

**`AssignPopover` becomes the shell.** The trigger, the pane that names itself for assistive tech, and content mounted only while open. It holds no choices and no commit.

**`AssignOptions` is that content** for a caller who already holds its choices: the title, the search, the inline suggestions, the in-flight latch, the failure shown in place. It stays in the kit deliberately, because the four assign surfaces must not drift apart in their labels, their empty states, or whether a failure is announced at all, which is exactly what happened to the two components `AssignPopover` originally replaced.

**`GroupAssignPicker` owns the two queries.** It is lazy because the shell gates mounting, which is the `FactionPicker`-inside-`FactionLoadPopover` shape, and AGENTS.md already names this exact case: a Picker "renders *through* the kit (a `Select`, an `AssignPopover`)".

**The group page keeps the one thing neither may own:** asking before a move. An asset already maintained by another group leaves it when added here, and that consequence is the page's knowledge. `onPick` resolving `false` leaves the pane open, which is how a reader backs out.

## The shell tells its content what it knows

`noun`, `title`, `disabled`, the label id and the close travel by context rather than props. A Picker mounted inside knows none of that and should not have to be told twice, so the words the shell chose and the words the contents use cannot disagree.

## Also removed

The group page's own `AssetAssignOption`, a second copy of the picker's item type.

## Verified

**The property the whole split rests on is now pinned.** A new kit test asserts the pane's content does not mount until it opens, and mounts fresh on each opening. Nothing asserted it before, and a story cannot: a story can only show what appears after a click, not that nothing existed before it.

Worth stating precisely, because I checked rather than assumed: **that test passes whether or not the shell gates on `opened` itself**, because Mantine's dropdown already declines to render its children while closed. So the assertion pins the *property*, not my line. I kept the explicit gate anyway and said why in the code: a Picker's read now depends on this, and one explicit line is better than an undocumented behaviour of a dependency. The test still catches what matters, a dropdown kept mounted or content moved outside it.

Full gate list: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 730 unit tests, 352 storybook tests.

## A rider: the arrow now points at its trigger

**Norbert caught this in the proof images below**, which is why it is here rather than parked. Pre-existing rather than introduced by this PR, folded in because this PR already rewrites the component, and filed as [#714](https://github.com/ndelangen/dunezone/issues/714).

The pane is 340px wide over a 40px icon trigger, so `shift` slides it to fit and the default side-anchored arrow stays pinned to the pane's corner, pointing at empty background.

Settled by measuring rather than reading, since the ticket asked for that. Arrow centre against trigger centre on the real story:

| viewport | before | after (`arrowPosition="center"`) |
|---|---|---|
| 400px | shifted hard | **0px off** |
| 520px | **75px off** | **0px off** |
| 1200px (unshifted) | side-anchored | **0px off** |

So the arrow stays rather than being dropped: it tracks the trigger exactly whether the pane is shifted or not, which is a better answer than the removal the ticket floated. **All four images below were recaptured with this fix in**, and that correction is why they moved.

## Proof

From `AssignPopover`'s own stories. All four assign surfaces render these same states.

| closed | open, with choices |
|---|---|
| trigger only, pane closed | pane open showing a faction to add |

| open, still fetching | open, nothing to offer |
|---|---|
| pane open showing Loading groups | pane open showing no groups available yet |

The loading state matters more after this change than before: a caller holding its choices was never loading, but a Picker inside the pane starts its read when the pane opens, so this is the first frame a reader now sees.

## Not verified, and I would rather say so

**The group detail page's signed-in path was not exercised in a browser.** The add-a-faction control only renders for an active member, and the local origin has no session; I will not handle credentials to get one.

So the deferral is argued structurally rather than observed on that page: the queries now live only inside `GroupAssignPicker`, the picker is only reachable as the shell's content, and the shell mounts content only while open, which the new test pins. A websocket-level check on the real page, the way [#702](https://github.com/ndelangen/dunezone/pull/702) was verified, is the thing that would close it, and it needs a signed-in member.

## Summary by CodeRabbit

* **New Features**
  * Added shared pickers for assigning owned factions and rulesets to groups.
  * Assignment pickers now show ownership locations, exclude items already in the group, and support search, loading, and empty states.
  * Assignment content loads when the picker opens, improving page responsiveness.

* **Improvements**
  * Standardized assignment popovers with reusable option panels, clearer error handling, and improved repositioning as content changes.

### DIFF
--- a/src/app/pickers/GroupAssignPicker.tsx
+++ b/src/app/pickers/GroupAssignPicker.tsx
@@ -1,0 +1,77 @@
+import { AssignOptions, useAssignPane } from '@ui/control/AssignPopover';
+
+import { useFactionsOwnedForGroupAssign } from '@db/factions';
+import { useRulesetsOwnedForGroupAssign } from '@db/rulesets';
+
+/** Enough of an owned asset to offer it and to say where it currently lives. */
+export type OwnedAssignItem = {
+  id: string;
+  name: string;
+  groupId: string | null;
+  groupName: string | null;
+};
+
+export type GroupAssignPickerProps = {
+  /** The group being added to, so an asset already in it is not offered. */
+  currentGroupId: string;
+  /**
+   * Reports the choice.
+   * The page decides what happens next, including whether to ask first: resolve `false` to leave the pane open, which is how a reader backs out of a move.
+   */
+  onPick: (item: OwnedAssignItem) => Promise<boolean | void>;
+};
+
+/**
+ * The presentation both owned-asset pickers share, once one of them has its rows.
+ * `noun` comes from the pane rather than a prop, so the words the shell chose and the words here cannot disagree.
+ */
+function OwnedAssignOptions({
+  items,
+  loading,
+  currentGroupId,
+  onPick,
+}: GroupAssignPickerProps & { items: OwnedAssignItem[]; loading: boolean }) {
+  const { noun } = useAssignPane();
+  const assignable = items.filter((item) => item.groupId !== currentGroupId);
+  const byId = new Map(assignable.map((item) => [item.id, item]));
+
+  return (
+    <AssignOptions
+      loading={loading}
+      searchLabel={`Search your ${noun}s`}
+      emptyMessage={
+        items.length === 0 ? `You don't own any ${noun}s yet.` : `All your ${noun}s are already in this group.`
+      }
+      options={assignable.map((item) => ({
+        value: item.id,
+        label: item.groupName ? `${item.name} — currently in ${item.groupName}` : `${item.name} — unassigned`,
+      }))}
+      onAssign={async (value) => {
+        const item = byId.get(value);
+        if (!item) {
+          return false;
+        }
+        return await onPick(item);
+      }}
+    />
+  );
+}
+
+/**
+ * The factions this viewer owns, offered for adding to a group.
+ *
+ * Mounted inside an `AssignPopover`, which mounts its content only while open, so the read starts when the reader signals intent and is torn down when they leave.
+ * Its query requires authentication, so only mount it for an active member.
+ * The owned lists are their own queries rather than part of the page bundle, decided on «Group detail's owned-asset lists» (#348/#182);
+ * what changed since is only when they are subscribed to.
+ */
+export function OwnedFactionAssignPicker(props: GroupAssignPickerProps) {
+  const owned = useFactionsOwnedForGroupAssign();
+  return <OwnedAssignOptions items={owned.data ?? []} loading={owned.isLoading} {...props} />;
+}
+
+/** The rulesets this viewer owns. Same rules as the faction picker: active members only. */
+export function OwnedRulesetAssignPicker(props: GroupAssignPickerProps) {
+  const owned = useRulesetsOwnedForGroupAssign();
+  return <OwnedAssignOptions items={owned.data ?? []} loading={owned.isLoading} {...props} />;
+}
--- a/src/app/routes/_app/assets/-assetEditorStates.tsx
+++ b/src/app/routes/_app/assets/-assetEditorStates.tsx
@@ -2,7 +2,7 @@ import { Alert, Anchor, Group, Stack, Text } from '@mantine/core';
 import { ASSET_TYPES, isAssetType } from '@shared/assets/types';
 import { Link, useNavigate } from '@tanstack/react-router';
 import { PageTitle } from '@ui/block/PageTitle';
-import { AssignPopover } from '@ui/control/AssignPopover';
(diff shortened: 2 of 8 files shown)


ndelangen (commented):
Orchestrator review at 8c34623, base main, branch at rest. Approved pending CI. The decision is executed exactly: the shell gates mounting, GroupAssignPicker owns both queries and is lazy by position, AssignOptions carries the presentation, and the pane context making the shell's chosen words binding on its content (noun from useAssignPane, so shell and picker cannot disagree) is a better mechanism than the decision required. Verified the finding is dead: the group page's four owned-for-assign references on main are zero on this branch. The confirmation staying with the page is the right ownership call, and the AssetAssignOption dedup rides naturally.

The test-weakness disclosure is accepted and the explicit gate kept for the stated reason is correct: a Picker's read now depends on closed-means-unmounted, and one written line beats an inherited framework behaviour nobody promised. The signed-in browser gap folds into the maintainer's existing post-deploy glance: the same signed-in session that checks #702's profile Select can open a group page's network panel and confirm no assign query fires on load. Both checks, one sitting.

ndelangen:
## Three defects found by an independent review, all fixed (second commit)

All three LLM reviewers were dark on this PR (CodeRabbit rate-limited, Copilot out of quota, Sourcery absent), so I ran an adversarial pass. Four lenses, every finding put to two refuters. Two findings killed, three survived — and **the first was found independently by four of the four lenses, with every refuter failing to refute it.**

They share one root cause, which is the new fact this PR introduces: **a Picker inside the pane starts its read on mount, so the contents can now arrive after the pane opens.** Two things assumed otherwise.

### The focus trap had nothing to hold

Mantine's `FocusTrap` runs `focusNode` once when the dropdown attaches, and does not re-run when the subtree changes. The search field was gated behind `!loading && options.length > 0`, so the first frame inside the dropdown was two `Text` nodes. The trap found nothing focusable, fell back to the dropdown `div` (which carries `tabIndex={-1}` from `withRoles`), and when the field mounted a moment later it never received the caret.

So on the group page: open the pane, type the faction's name, and nothing is filtered. The keystrokes go to the container. You must Tab or reach for the mouse first. On `main` the query was already resolved at page mount, so the field was there on the first frame and the trap focused it.

The field and the list are now present from the first frame, with the loading line inside the list rather than instead of it. That also means a reader can start typing while the choices arrive.

### The placement locked while the pane was short

Mantine 9.5.1 defaults `preventPositionChangeWhenVisible: true`: it measures the dropdown once visible, locks the placement, and **drops the `flip` middleware entirely**. Measured at ~90px the loading pane approved `bottom-start`; the options then arrived and it grew ~260px downward with no flip left to pull it back, and `shift` corrects x rather than y for that placement.

The placement stays live now, because a pane whose contents fetch is not settled when it opens.

### A deleted type left its docstring behind

`AssetAssignOption`'s JSDoc ended up on `RosterEntry`, describing something else entirely. Mine, from this PR.

### Verified

The focus fix is pinned by a test asserting the search field is present while the choices are still arriving. I checked it is not vacuous: it fails against the old `!loading && options.length > 0` shape.

Full gate list: lint, format:check, check:prose, typecheck, knip, 731 unit tests, 351 storybook tests.

**Worth saying plainly:** both real defects are in the signed-in group-page flow that this PR's body already stated was never exercised in a browser. The gap I flagged is exactly where the defects were.

ndelangen (commented):
Orchestrator post-merge concurrence on the delta 8c34623..15f4c6ae, which the merge outran: two rider commits, both verified. The arrow fix (arrowPosition center with the shift reasoning written in place) is confirmed by the recaptured proof image, arrow tip at the trigger's centre; that is the standard of proof the fix demanded. The pane-shape commit is a genuine catch riding along: Mantine's focus trap runs once at dropdown attach, so a pane that opens with nothing focusable leaves the caret on the dropdown itself and a reader types into nothing when the field arrives late; keeping field and list present from the first frame with loading shown inside the list, pinned by a new test, fixes an interaction defect the loading-placeholder swap concealed.

Issue #714, opened 2026-08-24
The assign pane's arrow points at the desk, not the trigger


ndelangen:
Flagged by Norbert from PR #713's proof image, and pre-existing rather than introduced there: the popover arrow sits at the pane's left corner roughly 150px from the trigger's center, pointing at empty background.

Proof image: https://github.com/ndelangen/dunezone/releases/download/proof-assets/pr-713-assign-open-with-options.png

Mechanism, read from AssignPopover.tsx: the Popover uses position="bottom-start" with withArrow and no arrowPosition, so the arrow takes Mantine's default corner placement relative to the dropdown. The pane is much wider than its trigger, and when the dropdown is slid to fit (or start-aligned while the trigger sits mid-pane-width), the arrow stays pinned to the dropdown's corner instead of tracking the trigger. Candidates, to be settled by trying rather than reading: Mantine's arrowPosition and arrowOffset props, or dropping the arrow entirely if no configuration tracks the trigger under a wide shifted pane; several of the app's other floating panes carry no arrow, so consistency may favor removal.

Small, visual, and worth a proof image of its own when fixed. Found because the proof rule put the pixels where the maintainer would see them, which is worth noting as the rule's third catch.

ndelangen:
Fixed in #713's rider commit at Norbert's request from the proof review: arrowPosition center with the mechanism documented inline, verified by the recaptured proof image showing the arrow tip at the trigger's centre. https://github.com/ndelangen/dunezone/releases/download/proof-assets/pr-713-assign-open-with-options.png

PR #718, opened 2026-08-24
The last two popovers point their arrow at the control that opened them

### LINKED ISSUE #715: Two more popovers park their arrow at the pane's corner
Found while fixing [#714](https://github.com/ndelangen/dunezone/issues/714). Same defect class, same one-word fix, filed separately rather than widening a PR that was already under review.

## The mechanism, measured

A `Popover` with `withArrow` and no `arrowPosition` anchors the arrow to the pane's own corner. When the pane is much wider than its trigger, `shift` slides it to fit and the arrow is left pointing at background rather than at the control that opened it.

On `AssignPopover` (340px pane, 40px icon trigger) I measured the arrow **75px** off the trigger's centre at a 520px viewport, growing as the pane is pushed further. `arrowPosition="center"` put it **0px off at 400, 520 and 1200px**, shifted or not.

## The two that still have it

Both are wide panes over narrow `IconAction` triggers, with `withArrow` and no `arrowPosition`:

| file | position | width |
|---|---|---|
| [`widgets/faction-editor/FactionLoadPopover.tsx:27`](src/app/widgets/faction-editor/FactionLoadPopover.tsx) | `bottom-start` | 440 |
| [`routes/_app/rulesets/$rulesetSlug/index.tsx:78`](src/app/routes/_app/rulesets/$rulesetSlug/index.tsx) | `bottom-end` | 440 |

The fix is `arrowPosition="center"` on each. I read their configuration rather than measuring these two, so someone should confirm the offset on each before and after rather than taking the mechanism on trust; `bottom-end` in particular I have not exercised.

## Worth noting

This is the arrow fix's own second-source problem: three popovers each spell out the same floating configuration by hand. If a fourth appears it will spell it out again. Whether the shared shape earns a wrapper is a separate question from fixing these two.

ndelangen:
Closes [#715](https://github.com/ndelangen/dunezone/issues/715). One word each, on the two popovers [#714](https://github.com/ndelangen/dunezone/issues/714) left behind.

## The mechanism

A `Popover` with `withArrow` and no `arrowPosition` anchors the arrow to the pane's own corner. A 440px pane over a 40px trigger gets slid by `shift` to fit the viewport, and the arrow goes with the pane rather than staying on the control that opened it.

| file | position | width |
|---|---|---|
| `widgets/faction-editor/FactionLoadPopover.tsx` | `bottom-start` | 440 |
| `routes/_app/rulesets/$rulesetSlug/index.tsx` | `bottom-end` | 440 |

`ControlBlock`'s help affordance also carries `withArrow`, but it is a `Tooltip`: narrow and following its target, so not this defect class. These two are the last of it.

## Measured, because the filing had not been

The ticket asked for this explicitly: it inspected these two components' configuration rather than measuring them, and said `bottom-end` in particular had never been exercised. That caution was worth having, because the placement turns out to matter more than the position does.

The variable is where the trigger sits, since that is what decides how far `shift` has to slide the pane. A toolbar can put its control anywhere, so this sweeps it rather than picking one:

| position | trigger | viewport | default | `arrowPosition="center"` |
|---|---|---|---|---|
| bottom-start | start | 560 / 900 / 1440 | -7 | **0** |
| bottom-start | middle | 560 | **-155** | **0** |
| bottom-start | middle | 900 / 1440 | -7 | **0** |
| bottom-start | end | 560 / 900 / 1440 | 8 | **0** |
| bottom-end | start | 560 / 900 / 1440 | -7 | **0** |
| bottom-end | middle | 560 | **156** | **0** |
| bottom-end | middle | 900 / 1440 | 8 | **0** |
| bottom-end | end | 560 / 900 / 1440 | 8 | **0** |

Arrow centre against trigger centre, in pixels. Across all eighteen combinations the default is off by up to **156px** and never better than 7px; `arrowPosition="center"` is **0px off in every one**.

The worst case is exactly the one that had gone unexercised. `bottom-end` is well behaved when its trigger sits at the right of a toolbar, which is what end-alignment is for, and falls apart when the trigger moves inward and the pane has to be shifted.

## Proof

`bottom-end`, trigger mid-toolbar, 560px viewport. The pane has been slid 156px away from the control that opened it.

| before | after |
|---|---|
| the arrow parked at the pane's far corner, pointing at background | the arrow directly beneath its trigger |

## How this was measured, and what that does not cover

Both popovers live behind a session. The faction loader is on the create and edit routes; the ruleset one renders only for a viewer with `capabilities.edit`. I will not handle credentials to reach them, so I did not photograph the pages themselves.

What I measured instead is the real Mantine `Popover` and the real `IconAction`, mounted in a scratch harness with the exact props each page passes, so the floating computation under measurement is the one those pages run. The harness is not committed; it existed to produce the table above.

That leaves one honest gap: the numbers are the floating behaviour these two components will get, not an observation of the two pages. What would close it is opening either control while signed in and reading the arrow's offset, which is a minute's work for anyone with a session.

I also got the harness wrong once and it is worth saying so, because the wrong version looked convincing. My first attempt put the trigger at the container's leading edge for both positions, which made `bottom-end` and `bottom-start` produce identical numbers: at that placement `shift` clamps them to the same box, so the harness reported the two positions as indistinguishable and the `bottom-end` case as barely affected. Placing the trigger where a toolbar actually puts it is what surfaced the 156px.

## Also worth noting, not done here

Three popovers now spell out the same floating configuration by hand, and a fourth would spell it out again. The ticket flags this and says whether the shared shape earns a wrapper is a separate question; it still is, and this PR does not answer it.

## Verified

Gates: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 731 unit tests, 361 storybook tests.

### DIFF
--- a/src/app/routes/_app/rulesets/$rulesetSlug/index.tsx
+++ b/src/app/routes/_app/rulesets/$rulesetSlug/index.tsx
@@ -82,6 +82,8 @@ function AddFactionPopover({
       width={440}
       shadow="md"
       withArrow
+      /* A 440px pane over a 40px trigger is slid by `shift` to fit, and a side-anchored arrow goes with the pane rather than staying on the control that opened it (#715). */
+      arrowPosition="center"
       trapFocus
       returnFocus
     >
--- a/src/app/widgets/faction-editor/FactionLoadPopover.tsx
+++ b/src/app/widgets/faction-editor/FactionLoadPopover.tsx
@@ -31,6 +31,8 @@ export function FactionLoadPopover({ disabled, currentPublicSlug, onLoaded }: Fa
       width={440}
       shadow="md"
       withArrow
+      /* A 440px pane over a 40px trigger is slid by `shift` to fit, and a side-anchored arrow goes with the pane rather than staying on the control that opened it (#715). */
+      arrowPosition="center"
       trapFocus
       returnFocus
     >


ndelangen (commented):
Orchestrator review at 1198eee7ae9, base main, branch at rest. Approved pending the settling checks. Two one-word fixes each carrying the constraint in a comment, backed by the eighteen-combination sweep: the default up to 156px off and never better than 7, arrowPosition center 0px off in every cell, with the worst case (bottom-end, mid-toolbar trigger) being exactly the placement the filing had never exercised. The harness near-miss is the valuable record: a trigger at the leading edge made shift clamp both positions to the same box and the wrong version was convincing; a harness that does not reproduce the page's layout measures its own layout. The stated coverage limit (component-level numbers with exact props, not a page-level observation, one signed-in minute to close) stays visible rather than dressed as a pass, which is the standard.

Issue #701, opened 2026-08-24
Work the composition wave


ndelangen:
Execute the composition wave, second of the three post-kit waves per the ordering decision on #697. Contents from the route sweep (#631), worked as small reviewed PRs:

1. The page-message widget: the pending, error, not-found and login-gate frames pasted across about seventeen sites (about 200 lines), with the assets editor's local extraction as the proven shape.
2. The entity-mention convergence (#685): fully specified on its ticket; four chip components, Links.Item's leading visual, fourteen sites.
3. The #451 identity-band territory: five hand-rolled detail-page bands, the drifted toolbar cluster, the remaining class-carrying titles, and the documented ink override inconsistency; read the scope corrections on #451 before starting.
4. The FAQ tag fieldset pasted verbatim across two routes.
5. Kit lookalikes: the groups page's hand-rolled Links, the rulesets index's hand-rolled OpenableTile and TileGrid, the FAQ pages' raw elements over kit answers.
6. The /auth TanStack starter copy.

Sequencing: starts when the data wave (#698) releases the worker or a second lane opens without file overlap; #685's conversions and item 3 both touch the profile and detail routes, so they do not run concurrently with data-wave items in those files.

ndelangen:
Constraint from #719, binding on item 1: LoadError (src/app/ui/block) is the load-failure BODY, words and affordance with no chrome; the page-message widget owns the frame (chrome, header, way back, one slot) and takes bodies as its slot content. It must not grow an error case of its own; the error body lives in LoadError, and the pending, not-found and login-gate kinds want sibling bodies beside it (the pending body is already written at least twice). AssetDetailMessage is the proven frame shape, already composing LoadError on the asset detail route.

ndelangen:
Item 6 is delivered: [A typed /auth lands on sign-in instead of scaffolding copy (#725)](https://github.com/ndelangen/dunezone/pull/725) is merged, verified by tip ancestry and the merged file. Item 1 (the page-message widget) is in flight on the helper's lane, slice 1 built and in visual pass. Items 2, 3 and 5 stay parked until item 1's shape lands; item 4 (the FAQ fieldset) is open for a free lane.

ndelangen:
[...]

PR #719, opened 2026-08-24
A route's load failure offers the way forward it was already entitled to

### LINKED ISSUE #700: Five route error components swallow the stale-tab refresh prompt
Found while reviewing the catalogue narrowing ([#699](https://github.com/ndelangen/dunezone/pull/699)). Not introduced by it: the behaviour is the same on `main`. Filing rather than folding in, because it spans five routes and only two of them are touched by that PR.

## What happens

`StaleClientDataError` exists so that a tab running an older bundle, receiving data that no longer matches its compiled expectations, gets a way forward instead of a crash. `AppErrorComponent` in [`router.tsx`](src/app/router.tsx) honours that: it detects the error and renders "This page changed" with a **Refresh** button.

Five routes override `errorComponent`, and none of them detect it:

| route | error component |
|---|---|
| `_app/index.tsx` | `HomepageError` |
| `_app/factions/index.tsx` | `FactionCatalogueError` |
| `_app/factions/$factionId/index.tsx` | `FactionDetailError` |
| `_app/rulesets/$rulesetSlug/index.tsx` | `RulesetDetailError` |
| `_app/assets/$type/$slug/index.tsx` | `AssetDetailError` |

TanStack Router resolves `route.options.errorComponent` ahead of `defaultErrorComponent`, so on these five the stale case renders as a plain red alert with `error.message` and no refresh affordance. The reader is told "Faction data does not match this version of the app" and left to work out that reloading fixes it.

Routes that do not override, `/profiles/$profileSlug` among them, behave correctly.

## Why it matters more now

Any narrowing of a client-boundary contract makes this path reachable on purpose, for the length of a deploy. The catalogue narrowing does exactly that on two of these five routes. The mechanism is doing its job; the message just arrives without its button.

## Shape of the fix

The stale case wants one owner rather than five copies of an `instanceof` check. Worth noting that `AppErrorComponent` is deliberately inline-styled because it also covers routes outside `MantineProvider`, so a shared piece either has to work in both contexts or the two contexts need separate presentation over one shared predicate.

### LINKED ISSUE #704: Deleting a shell-held public Convex query errors the whole site for the deploy window
Found by an independent review of [#702](https://github.com/ndelangen/dunezone/pull/702) after it merged. Not a bug in that PR's code, and not something to fix in a file: it is a property of the deploy pipeline that makes one class of change unsafe, and #702 was the first change of that class.

## The window

[`deploy-main.yml`](.github/workflows/deploy-main.yml) ships the Convex functions **first** and the browser bundle **last**:

| step | line |
|---|---|
| Deploy Convex | 34 |
| Run and verify required migrations | 40 |
| Build unified Worker release | 74 |
| **Deploy exact Worker release** | 91 |

The Worker serves the site's JS. So between those two steps, every visitor downloads the **previous** bundle while the backend already runs the **new** functions.

If the change **deleted a public query the shell holds**, that old bundle subscribes to a function that no longer exists. Convex answers "Could not find public function", `convex/react`'s `useQuery` rethrows during render, and because `useCurrentProfile` is held by `SiteNavigation` inside the `/_app` layout, the throw lands on the router's `defaultErrorComponent`. Neither `src/app/routes/_app.tsx` nor `src/app/routes/__root.tsx` defines an `errorComponent`, so nothing narrower catches it.

The result is a **full-page error on every `_app` page**, for every visitor, signed in or out, for the length of the window. Tabs already open break at the instant the Convex deploy finishes.

This is a harsher failure than the stale-tab case. `StaleClientDataError` produces a refresh prompt because the data merely fails to parse. A missing function is a hard query error with no such affordance.

## How wide, measured

On [the #702 deploy](https://github.com/ndelangen/dunezone/actions/runs/32717958994):

| step | finished |
|---|---|
| Deploy Convex | 10:40:44 |
| Deploy exact Worker release | 10:41:42 |

**58 seconds.** Narrow, and there is no sign anyone hit it.

The reason to file this anyway is that the width is not bounded by anything relevant. `migrations:deploy` sits inside the window and carries a guard budget of 2,700,000 ms (`package.json`) — **45 minutes**. That deploy happened to run no migrations. A change that deletes a shell-held query *and* carries a migration could hold the site on an error page for as long as the migration takes.

## The rule this suggests

Deleting a public Convex query that any always-mounted component holds is an expand/contract change, not an edit:

1. Add the new query. Move the client to it. Ship.
2. Delete the old query in a **later** release, once no deployed bundle asks for it.

For #702 that would have meant keeping `profiles.currentUserId` and `profiles.current` for one release as thin re-exports over the same reads.

Worth deciding whether this belongs in AGENTS.md beside the other client-boundary rules, since the same pipeline ordering will apply to every future query rename or removal. Item 3 of #698 (the assign-popover deferral) does not delete any query, so it is unaffected.

ndelangen:
Closes [#700](https://github.com/ndelangen/dunezone/issues/700). Closes [#704](https://github.com/ndelangen/dunezone/issues/704).

Two deploy-window behaviours, argued separately because they are separate, in one PR because they sit on the same surface and reviewing them apart means reading it twice.

---

# #700, the swallowed refresh prompt

`StaleClientDataError` exists so a tab running an older bundle, sent data that no longer matches what it was compiled against, gets a way forward instead of a crash. The router's error component honours it: "This page changed", and a Refresh button.

Five routes override `errorComponent`, and none of them asked the question. TanStack resolves a route's own `errorComponent` ahead of the default, so on the homepage, the faction catalogue, the faction detail, the ruleset detail and the asset detail, the stale case arrived as a red alert reading "Faction data does not match this version of the app" with nothing to click. The mechanism was working; the message was simply losing its button.

## One owner for the question, two answers

The question is now `isStaleClientData`, beside the error it asks about.

Two answers rather than one, because the two callers cannot share a presentation. The router's component also covers routes outside `MantineProvider` and styles itself inline; a route's own component renders inside it. That is the split [#700](https://github.com/ndelangen/dunezone/issues/700) itself predicted.

Inside Mantine, `LoadError` is `FormError`'s sibling and composes it: an action that did not happen and a page that did not load are the same sentence with the same treatment, so the colour and the live region stay stated once. What it adds is the second case, and the fallback the five routes each wrote by hand for an error carrying no message.

It is a kit Block and not a Widget, which is worth saying because I built it as a Widget first and the lint told me otherwise: widgets may not import a data-layer value either. That is the taxonomy working rather than an obstacle. The page classifies the error, the component renders it, and `stale` arrives as a prop.

`StaleClientDataError` is no longer exported. The throw is its module's and so is the question about it, and knip noticed the moment nothing matched the class by name any more.

## Proof

The stale condition is **staged**: the catalogue's client schema is given a field no stored row has, which is precisely what a narrowing deploy does to a tab running the previous bundle. Identical staging on both sides, so the pair compares like with like, and reverted rather than committed.

Both sides are real code. The before is `main` with the app restarted against it, and I checked which tree the dev server was serving each way.

| on `main` | this branch |
|---|---|
| a red alert saying faction data does not match this version of the app, with nothing to click | this page changed, with a refresh button |

Read off the same renders:

| | alert text | refresh button |
|---|---|---|
| on `main` | "Faction catalogue could not be loaded / Faction data does not match this version of the app" | **absent** |
| this branch | "This page changed / The data no longer matches this version of the app." | **present** |

The page keeps its own chrome either way, since a header and a way back are that route's to choose.

---

# #704, the deleted shell-held query

This one is not a bug in a file, and the change is a rule rather than code.

Production ships the Convex functions first and the browser bundle last, so between those two steps every visitor downloads the *previous* bundle against the *new* functions. That order is right for adding a function and wrong for removing one: an old bundle subscribing to a deleted query gets "Could not find public function", `useQuery` rethrows during render, and a query held by an always-mounted component takes every `_app` page down with it.

Both of the issue's load-bearing claims check out, and I verified them rather than quoting them: the pipeline order is as filed, and `migrations:deploy` carries a **2,700,000 ms** guard budget, so the 45 minutes is real. The measured window was 58 seconds only because that deploy ran no migrations.

The rule now sits beside the other Convex doorway rules in `AGENTS.md`: add the new query, move the client, ship, and delete the old one in a later release.

## What I deliberately did not build

A safety net on `_app` or `__root`, which is the obvious next thought and which I think is wrong here.

The affordance such a component would offer is a refresh, and inside the window a refresh does not help: the Worker is still serving the bundle that asks for the missing function, so it loops until the Worker deploy lands. A message that named the deploy instead would have to guess, and would mislabel every unrelated shell error as a deploy in progress. Detecting the specific Convex wording would make a second source out of someone else's error string.

The asymmetry that causes this cannot be fixed by reordering either, since deploying the Worker first would break additive changes instead. The release discipline is the fix, which is what the issue argued and what I am agreeing with rather than adding to.

Say the word if you want the net anyway and I will build it, but I would rather not ship an affordance that does not work.

---

## Where this sits against the composition wave

Asked in review, and worth answering in the body because the wave should build on this rather than beside it.

[#701](https://github.com/ndelangen/dunezone/issues/701) item 1 is the page-message widget: the pending, error, not-found and login-gate **frames** pasted across about seventeen sites, with the assets editor's local extraction as the proven shape.

`LoadError` is not that widget and does not overlap it. The line between them is chrome and words:

- the **frame** owns the page around a message, a header, a way back, and a slot;
- the **body** owns what goes in the slot for one kind of message, its words and its affordance.

`LoadError` is the body for the error kind, and only that. It renders no chrome, which is why all five routes still supply their own.

This PR already shows the two composing, because one of the five routes had extracted its frame already. `AssetDetailMessage` in the asset detail route is exactly #701's proven shape: chrome plus a slot, holding three different bodies for loading, absent and failed. Its error case now reads:

```tsx
<AssetDetailMessage>
  <LoadError title="This asset could not be loaded" stale={isStaleClientData(error)}>{error.message}</LoadError>
</AssetDetailMessage>
```

So the answer is: deliberately the error kind's body, designed to be held by the frame the wave builds, not a first slice of the frame itself.

What that asks of the wave, concretely: when the page-message widget lands it should take the body as a slot and leave the error body here, rather than growing an error case of its own. The other three kinds want bodies beside this one, and the pending body is already written twice in this repo, in `AssetDetailPending` and wherever the other sites paste it.

I did not design this against #701, to be clear. It came out as the body because the chrome differs per route and the words do not. It happens to land on the right side of that line, and the constraint above is what keeps it there.

## About the `AGENTS.md` change

Operational documentation rather than a taxonomy amendment: a release rule with a measured window, filed beside the other Convex doorway rules. It changes no category, no boundary and no lint.

---

## Verified

Gates: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 731 unit tests, 364 storybook tests, and `publisher:release:verify` clean with no manifest diff.

`LoadError` carries three stories, including the stale state and an error with no message of its own, because both are states nobody sees by accident.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -376,6 +376,25 @@ Convex seam suites sit beside the module they cover and are named `<domain>.<fac
 (`factions.catalogue.test.ts`, `groups.softDeletion.test.ts`). One facet per file, so a suite's name
 says what broke before the failure text does.
 
+**Deleting a public query the shell holds is an expand/contract change, not an edit.** Production ships
+the Convex functions first and the browser bundle last
+([`deploy-main.yml`](.github/workflows/deploy-main.yml)), so between those two steps every visitor
+downloads the *previous* bundle against the *new* functions. That order is right for adding a
+function and wrong for removing one: an old bundle subscribing to a deleted query gets "Could not
+find public function", `useQuery` rethrows during render, and a query held by an always-mounted
+component takes every page in `_app` down with it. Neither `routes/_app.tsx` nor `routes/__root.tsx`
+defines an `errorComponent`, and refreshing does not help, because the Worker is still serving the
+bundle that asks for the missing function. So:
+
+1. Add the new query, move the client to it, ship.
+2. Delete the old one in a **later** release, once no deployed bundle asks for it.
+
+The window is not as narrow as it looks. It measured 58 seconds on
+[the #702 deploy](https://github.com/ndelangen/dunezone/actions/runs/32717958994), which ran no
+migrations, and `migrations:deploy` sits inside it carrying a 45-minute guard budget. A deletion that
+also carries a migration can hold the site on an error page for as long as the migration takes. This
+is harsher than the stale-tab case, which at least ends in a refresh prompt.
+
 Moving a file named in `RENDERER_RUNTIME_CLOSURE_PATHS` changes the renderer digest, so
 `publisher:release:verify` will report a `renderer-manifest.generated.ts` diff to commit. That is
 identity, not staleness: regeneration is driven by the checked-in `renderer_revisions`, which a move
--- a/src/app/db/core/clientBoundary.ts
+++ b/src/app/db/core/clientBoundary.ts
@@ -2,13 +2,25 @@ import type { z } from 'zod';
 
 /**
  * Incoming data no longer matches this bundle's compiled expectations, the stale-tab signal.
- * Rendered by the router's error component as a refresh prompt instead of a crash.
+ * Rendered as a refresh prompt instead of a crash.
+ *
+ * Not exported: the throw is this module's and so is the question about it, so callers ask `isStaleClientData` rather than matching the class themselves.
  */
-export class StaleClientDataError extends Error {
+class StaleClientDataError extends Error {
   constructor(context: string, options?: { cause?: unknown }) {
     super(`${context} does not match this version of the app`, options);
     this.name = 'StaleClientDataError';
   }
+}
+
+/**
+ * Whether a caught error is the stale-tab signal, and so wants a refresh rather than a red alert.
+ *
+ * One owner for the question, because it is asked from two places that cannot share an answer's presentation: the router's error component renders outside `MantineProvider` and styles itself inline, while a route's own error component renders inside it.
+ * Five route error components used to skip the question entirely and show the message with no way forward (#700).
+ */
+export function isStaleClientData(error: unknown): boolean {
+  return error instanceof StaleClientDataError;
 }
 
 /** Validate data entering the client runtime against this bundle's schema. */
--- a/src/app/router.tsx
+++ b/src/app/router.tsx
@@ -1,11 +1,11 @@
 import { ErrorComponent, createRouter as createTanStackRouter } from '@tanstack/react-router';
 
-import { StaleClientDataError } from '@app/db/core/clientBoundary';
+import { isStaleClientData } from '@app/db/core/clientBoundary';
 
 import { routeTree } from './routeTree.gen';
 
 function AppErrorComponent({ error }: { error: Error }) {
-  if (error instanceof StaleClientDataError) {
+  if (isStaleClientData(error)) {
     return (
       <div role="alert" style={{ maxWidth: '28rem', margin: '4rem auto', textAlign: 'center' }}>
         <h1>This page changed</h1>
--- a/src/app/routes/_app/assets/$type/$slug/index.tsx
+++ b/src/app/routes/_app/assets/$type/$slug/index.tsx
@@ -15,6 +15,7 @@ import { RULESET_ASSET_SLOTS } from '@shared/rulesets/assetSlots';
 import type { RulesetAssetSlot } from '@shared/rulesets/assetSlots';
 import type { ErrorComponentProps } from '@tanstack/react-router';
 import { createFileRoute, Link, notFound } from '@tanstack/react-router';
+import { LoadError } from '@ui/block/LoadError';
 import { OpenableTile } from '@ui/block/OpenableTile';
 import { PageTitle } from '@ui/block/PageTitle';
 import { Section } from '@ui/block/Section';
@@ -47,6 +48,7 @@ import type { ReactNode } from 'react';
 
 import { loadAssetPage, useAssetPage } from '@app/db/assets';
 import type { AssetPageData } from '@app/db/assets';
+import { isStaleClientData } from '@app/db/core/clientBoundary';
 import { AssetFace } from '@app/widgets/asset-face/AssetFace';
 
 import { useAssetDeletion, useAssetGroupActions } from '../../-assetEditorStates';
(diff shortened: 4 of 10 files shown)


ndelangen (commented):
Orchestrator review at adbbb76c60a, base main, branch at rest, everything green except the Codacy false positive awaiting Norbert's ignore (the eleven-digit Actions run id in a documentation link read as a phone number; the URL is the 58-second measurement's evidence and stays as written). Approved.

Verified in the diff: the class un-exports behind isStaleClientData so callers ask the module rather than matching the class (the question stays with the throw), the five routes' error components detect the stale signal and render the refresh path they previously swallowed (#700), the AGENTS.md expand/contract entry rides as operational documentation with a measured window (the #704 half: the rule that prevents the deletion class rather than code that cannot), and LoadError lands as the message BODY with no chrome, composing into AssetDetailMessage exactly as #701 item 1 will. The Sonar decline (zero Readonly occurrences across the kit; one file adopting it would introduce the inconsistency the rule claims to prevent) and the body-versus-frame answer are both on the record. #701 is annotated with the constraint: the page-message widget takes the body as a slot and must not grow an error case of its own.

PR #721, opened 2026-08-24
The shell stops writing a route attribute nothing reads

ndelangen:
First of the three AppRoot items on [#707](https://github.com/ndelangen/dunezone/issues/707), split out because it is the only one that needs no decision. The other two are [proposed on the issue](https://github.com/ndelangen/dunezone/issues/707#issuecomment-5399971378).

## `data-route` had no reader

Not a stylesheet, not the shell, not an e2e spec. A repo-wide search for `data-route` and `dataset.route` finds it in exactly two places: where `AppRoot` writes it, and where `AppRoot.test.tsx` asserts that it did.

A write-only document attribute fails the census's defence trivially, so it goes.

## The assertion goes with it

`AppRoot.test.tsx` asserted the attribute on mount and its absence after unmount. With nothing reading the attribute, that test's only subject was that the write happened.

That is worth naming rather than quietly deleting: a test whose subject is an implementation detail keeps dead state alive, because removing the state then looks like a regression. The two assertions covering `data-initial-animate` and `--scroll-pct` stay, since both have real readers in `page.css`.

## What stays, and why the dependency now says so

The effect stays, and so does its `pathname` dependency, because the other attribute it writes is the point. `data-initial-animate` re-arms the backdrop's eased arrival on every navigation and carries no route of its own, so `pathname` is the trigger rather than a value the effect reads.

That is now stated in the file. An unused-looking dependency is exactly the thing a later reader tidies away, and doing so would silently stop the attribute being re-stamped on navigation.

## Verified

Gates: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 731 unit tests, 361 storybook tests, and `publisher:release:verify` clean with no manifest diff.

No behaviour changes, so there is nothing to photograph: the attribute this removes was never read by anything that renders.

### DIFF
--- a/src/app/shell/AppRoot.test.tsx
+++ b/src/app/shell/AppRoot.test.tsx
@@ -101,7 +101,7 @@ describe('AppRoot page header', () => {
     act(() => root.unmount());
   });
 
-  it('releases its document-level route and scroll state on unmount', () => {
+  it('releases its document-level page effects on unmount', () => {
     const container = document.createElement('div');
     const root = createRoot(container);
 
--- a/src/app/shell/AppRoot.test.tsx
+++ b/src/app/shell/AppRoot.test.tsx
@@ -115,13 +115,11 @@ describe('AppRoot page header', () => {
       );
     });
 
-    expect(document.documentElement.dataset.route).toBe('/privacy');
     expect(document.documentElement.hasAttribute('data-initial-animate')).toBe(true);
     expect(document.documentElement.style.getPropertyValue('--scroll-pct')).not.toBe('');
 
     act(() => root.unmount());
 
-    expect(document.documentElement.hasAttribute('data-route')).toBe(false);
     expect(document.documentElement.hasAttribute('data-initial-animate')).toBe(false);
     expect(document.documentElement.style.getPropertyValue('--scroll-pct')).toBe('');
   });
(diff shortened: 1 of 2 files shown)


ndelangen:
Orchestrator review at cab0613140466e5f1c7a93cc18722dfd8398d702: **concur, ready for Norbert's merge**. Single commit, branch at rest, all gates green including e2e_docker and a clean publisher verify.

Premise verified independently: a repo-wide `git grep 'data-route\|dataset\.route'` finds zero readers outside AppRoot.tsx and the write-assertions this PR deletes — the attribute was write-only. The diff is exactly scoped, and keeping the `pathname` dependency is correct: it is load-bearing for re-arming `data-initial-animate` per navigation, and the comment defending it says so accurately.

One optional nit, non-blocking: the enclosing test at AppRoot.test.tsx:104 is still titled 'releases its document-level route and scroll state on unmount' — its primary referent, `dataset.route`, no longer exists in the body. If the rename is taken, this note lapses with the new head and I re-anchor.

ndelangen:
Re-anchored orchestrator review at 103626b60eb9824770bbcae6c55fc8b43728c374: **concur, ready for Norbert's merge**. This head is exactly the previously reviewed cab0613140466e5f1c7a93cc18722dfd8398d702 plus one commit changing one line — the test title rename from the earlier nit ('releases its document-level page effects on unmount', now accurate). All gates green on this head including e2e_docker. Everything in the earlier review stands; nothing else changed.

Issue #722, opened 2026-08-24
The motion verdict reaches the first frame


ndelangen:
## Question

Resolved by decision on [the census](https://github.com/ndelangen/dunezone/issues/707#issuecomment-5400638786): the profile's motion toggle must beat first paint. What remains is the build.

A pre-hydration twin of `src/app/styles/motion.ts` joins the colour-scheme script in `src/app/routes/_app.tsx`'s head scripts: read the `motion` cookie, fall back to `prefers-reduced-motion`, stamp `data-motion` on `<html>` before first paint. AppRoot's existing effect keeps owning updates after hydration; the script owns only the first frames. Today the stylesheet falls back to the OS hint until React hydrates, so a motion-off visitor gets spinning dice on first paint and an OS-quieted opt-in visitor gets still ones.

Acceptance: the stamped attribute matches `motionAllowed()` for all four cookie-by-OS combinations before hydration, proven in the PR; both page.css fallback blocks stay (they still cover the no-JS case).

PR #723, opened 2026-08-24
The backdrop keeps its parallax and loses an arrival nobody saw

ndelangen:
Closes the scroll half of [#707](https://github.com/ndelangen/dunezone/issues/707), [decided by Norbert](https://github.com/ndelangen/dunezone/issues/707#issuecomment-5400638786) after both halves were measured. The motion half is #722, built separately.

## The decision, and what each half of it rests on

**The parallax keeps its JavaScript.** The scroll listener writing `--scroll-pct` stays, because CSS cannot carry it everywhere. Measured in all three engines with a page that actually scrolls:

| engine | `CSS.supports('animation-timeline', 'scroll()')` | backdrop pans |
|---|---|---|
| Chromium 151 | yes | yes |
| **Firefox 153** | **no** | **no** |
| WebKit 26.5 | yes | yes |

Mozilla's own [experimental features page](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features) has `layout.css.scroll-driven-animations.enabled` on by default in Nightly since 136 and **off by default in Release since 110**, with no announced target. That is why this is unlike CanvasScale, where the gap was eight days against a named version.

**The eased arrival goes**, because measurement showed it almost never runs:

| scenario | position changes | |
|---|---|---|
| navigate from a scrolled page | **1** | a jump. The navigation's own scroll stripped the attribute at 237ms; the position changed at 260ms |
| navigate from an unscrolled page | **0** | nothing to animate |
| first load on `/privacy`, `/terms`, `/` | **0** | nothing to animate |
| first load, page shorter than the viewport | **49 over 1194ms** | the only place it fired |

So `data-initial-animate`, the pathname effect that re-armed it on every navigation, and the 1.2s transition rule in `page.css` are all removed.

**The accepted cost, stated plainly:** on a page shorter than the viewport the backdrop now settles at its end position instantly instead of drifting there over 1.2s. That is a real if narrow loss, and it is what Norbert weighed against carrying the machinery.

## Removing it emptied a chain

The scroll handler existed only to wrap the attribute write, so the listener now takes `scheduleUpdate` directly.

`pathname` had no consumer once the effect went. It is gone from `AppRoot`, from `ApplicationChrome` which only forwarded it, and from both of that component's callers, taking a `useLocation` subscription with it in each. `AppNotFound` was the one the type error found rather than the trace, which is the trace's fault and worth admitting.

## Proof

Re-run of the same rig, against this branch. I verified the dev server was serving this branch by its bytes, because another session's vite from the root checkout held port 3000 and would have answered for `main`.

| check | result |
|---|---|
| parallax tracks the wheel | `50% 0%` → `24.99%` → `50.02%` → `75.01%` → `50% 100%`, **5 of 5 distinct** |
| the ease, on the one page that used to show it | **1 step, was 49 over 1194ms**; `data-initial-animate` absent; settles at `50% 100%` |

The first row is the behaviour the JS is kept for. The second is both the removal working and the accepted cost visible: it still arrives at the end position, it just no longer drifts.

## A note on the test

The unmount test's title is narrowed rather than left overstating. It covered `data-initial-animate` and `--scroll-pct`; with the first gone it covers one thing, so it says so.

`data-motion`'s release on unmount is covered by no test. Worth knowing, and it belongs to #722 rather than here, since that ticket is actively editing the motion path and a second writer in this file would only conflict.

## Verified

Gates: lint, format:check, check:prose, typecheck, knip, check:css-orphans, check:app-layout, 732 unit tests, 364 storybook tests, and `publisher:release:verify` clean with no manifest diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved page transition behavior by removing route-dependent animation handling.
  * Preserved scroll-progress updates and cleanup when navigating or unmounting.
  * Ensured the application shell renders consistently across routes and not-found pages.

### DIFF
--- a/src/app/routes/_app.tsx
+++ b/src/app/routes/_app.tsx
@@ -1,4 +1,4 @@
-import { createFileRoute, Outlet, useLocation } from '@tanstack/react-router';
+import { createFileRoute, Outlet } from '@tanstack/react-router';
 
 import { ApplicationChrome } from '@app/shell/ApplicationChrome';
 import { AppNotFound } from '@app/shell/AppNotFound';
@@ -24,10 +24,8 @@ export const Route = createFileRoute('/_app')({
 });
 
 function AppLayout() {
-  const pathname = useLocation({ select: (location) => location.pathname });
-
   return (
-    <ApplicationChrome pathname={pathname}>
+    <ApplicationChrome>
       <Outlet />
     </ApplicationChrome>
   );
--- a/src/app/shell/AppNotFound.tsx
+++ b/src/app/shell/AppNotFound.tsx
@@ -1,14 +1,12 @@
-import { Link, useLocation } from '@tanstack/react-router';
+import { Link } from '@tanstack/react-router';
 import { PageTitle } from '@ui/block/PageTitle';
 import { PageLayout } from '@ui/layout/PageLayout';
 
 import { ApplicationChrome } from './ApplicationChrome';
 
 export function AppNotFound() {
-  const pathname = useLocation({ select: (location) => location.pathname });
-
   return (
-    <ApplicationChrome pathname={pathname}>
+    <ApplicationChrome>
       <PageLayout>
         <PageLayout.Header>
           <PageTitle title="404 - Page Not Found" />
--- a/src/app/shell/AppRoot.stories.tsx
+++ b/src/app/shell/AppRoot.stories.tsx
@@ -16,7 +16,6 @@ const meta = preview.meta({
     },
   },
   args: {
-    pathname: '/factions',
     children: shellPageOptionLabels[0],
   },
   argTypes: {
--- a/src/app/shell/AppRoot.test.tsx
+++ b/src/app/shell/AppRoot.test.tsx
@@ -64,7 +64,7 @@ describe('AppRoot page header', () => {
     act(() => {
       root.render(
         chrome(
-          <AppRoot pathname="/privacy">
+          <AppRoot>
             <PageLayout>
               <PageLayout.Header>
                 <h1>Privacy policy</h1>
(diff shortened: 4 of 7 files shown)


ndelangen:
Orchestrator review at 0139e92d79ab5bc3718fbe26efddc08c7e56a5ae: **one finding, then ready**.

The removal itself is verified clean: zero `initial-animate` references anywhere on the head, the collapsed scroll handler registers and removes the same `scheduleUpdate` reference, and the dead `pathname` was chased through both `ApplicationChrome` callers with their `useLocation` subscriptions. The narrowed test title is accurate to what the test now asserts.

**Finding: `AppRoot.stories.tsx:19` still passes `pathname: '/factions'` in the meta args** — a prop that no longer exists on `AppRootProps`. It's inert at runtime (AppRoot destructures, nothing spreads to the DOM), which is why 361 storybook tests stayed green, and `preview.meta()`'s factory typing doesn't excess-property-check args, which is why typecheck stayed green. But a story is supposed to be correct usage on display, and this one now documents a prop the component doesn't take. One-line removal.

Worth keeping for the map: prop removals have a gate blind spot — story args pass both typecheck and the storybook run when they carry a deleted prop, so the sweep has to be manual (`git grep` the prop name after removing it).

This review lapses when the head moves; push the fix and I re-anchor. e2e_docker still pending as I write.

ndelangen:
Re-anchored orchestrator review at f1e3fdbef3ddede7e0a9b78638e115c00b60d980: **concur, ready for Norbert's merge**. This head is exactly the previously reviewed 0139e92d79ab5bc3718fbe26efddc08c7e56a5ae plus one commit removing the stale `pathname` story arg — the review's one finding, resolved. All gates green on this head. The helper's repo-wide grep confirmed both removed names return nothing anywhere (`pathname` gone from src/app, `data-initial-animate` gone from the repo).

PR #726, opened 2026-08-24
A page with nothing to show wears one frame, and says which nothing it is

ndelangen:
First slice of item 1 on [#701](https://github.com/ndelangen/dunezone/issues/701): the page-message widget, its bodies, and three routes adopting it. The frame and body split is the [constraint Norbert set on the ticket](https://github.com/ndelangen/dunezone/issues/701#issuecomment-5399892600) after #719, and the site inventory is finding 1 plus finding 3 of the route sweep (#631).

The remaining sites come in two follow-ups, so the shape gets reviewed before it is installed fourteen more times.

## What owns what

`PageMessage` owns everything that is not the words: the layout, the band and its name, the pane, and the way back. Four blocks own the words, one per state, and the caller picks which one to hand over. That is the whole membrane.

| | what it says | what it owns beyond the words |
|---|---|---|
| `LoadPending` | still loading | a `status` live region, so a reader who cannot see the page hears that it is loading |
| `NotAvailable` | not there, or not yours | nothing announced: nothing failed and nothing is in flight |
| `LoadError` (already existed) | failed to load | the alert colour, the live region, the stale-tab reload |
| `LoginGate` | not signed in | the sentence around the caller's verb phrase, and the one destination |

The way back is `PageMessage.Back`, a `createLink` over a themed `Anchor`. It is the same idiom `Links.Item` already uses, so the destination stays router-typed at the call site while the anchor treatment stays with the frame. Nine of the sites this replaces spelled that anchor as the browser's default blue one.

## Two calls worth arguing with

**Three bodies, not four.** Norbert's constraint names pending, not-found and login-gate as three kinds beside `LoadError`. Not-found and access-denied render identically and differ only in words, so I folded them into `NotAvailable` rather than ship two components answering one question. The sweep groups them the same way ("pending / error / not found / access denied"). I read "kinds" as reader-facing states rather than a component count, and if that reading is wrong the split is one file.

**The pane moved from the body to the frame.** `AssetDetailMessage`, the shape the ticket calls proven, does not paint; its pending child paints a `Surface` of its own. I inverted that, because only Surfaces paint, and a Block rendering its own `Surface` would be a Block painting a pane. `FormError`'s JSDoc is careful to justify painting only because its colour is content rather than a pane, which is the exception that fixes the rule.

That inversion is visible, and it is the one thing in this PR a reader would notice.

## The visible change

A `LoadError` inside a message frame used to float on the page background. It now sits on the pane the frame paints, like every other message body.

Measured on `/rulesets/dreamrules` against the worktree dev server, with the route's loader made to throw. Same server, same data, same viewport; only the route file differs between the rows.

| | holder of the alert | painted | alert width at 1280 | alert width at 375 | inset |
|---|---|---|---|---|---|
| before (`10b19d1ab24`) | `main` | no | 1200px | 351px | 0px |
| after | the frame's `Surface` | yes | 1134px | 285px | 33px |

Neither width overflows its viewport. The mobile row is the interesting one: at 375px the alert used to run to the content edge with no side gutter, and now it keeps the pane's 32px inset like the pending and absent bodies always did.

To see it rather than take my word: check out `10b19d1ab24`, add `throw new Error('x')` as the first line of the ruleset detail loader, and open `/rulesets/dreamrules`. Then do the same on this branch.

## Verified in the browser, not only in the gates

Dev server run from this worktree on 3207, confirmed serving this branch by its bytes rather than by which directory I thought it was in. Every state below was reached by typing a URL, except the pending row, which needs a client-side navigation and a slow loader because a cold load blocks before React mounts anything.

| state | route | outline | live region | way back | on a pane |
|---|---|---|---|---|---|
| absent | `/rulesets/there-is-no-such-ruleset` | `h1` Ruleset, `h2` Ruleset not found | none, correctly | Back to rulesets | yes |
| absent | `/assets/token-disc/there-is-no-such-token` | `h1` Disc tokens, `h2` Asset not found | none, correctly | Back to disc tokens | yes |
| failed | `/factions/there-is-no-such-faction` | `h1` Faction | `alert` | Back to factions | yes |
| pending | `/rulesets/dreamrules`, loader slowed | `h1` Ruleset, `h2` Loading ruleset | `status` | Back to rulesets | yes |

The asset row is worth reading twice: the per-type words work, so the frame says "Disc tokens" and offers "Back to disc tokens" rather than a generic label. The pending row's live region is new behaviour. Five routes wrote that heading and sentence by hand and none of them announced it.

## Adopted

Three routes, chosen to cover all four bodies between them rather than to be easy:

- `assets/$type/$slug/index.tsx` keeps its local `AssetDetailMessage`, which now supplies only the per-type words. That is the split working: the route knows the type, the widget knows the frame.
- `rulesets/$rulesetSlug/index.tsx`, all three frames.
- `factions/$factionId/index.tsx`, including the verbatim inline copy of its own pending component that #631 called out. It is now a call to that component.

No copy changed on any of them. The three routes lose 124 lines and gain 49. The four components and their stories cost 239, which the remaining fourteen sites are what pays back.

## Verified

Local gates: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 732 unit tests, 373 storybook tests (was 364; nine new stories).

`bun run check` reports "no capture-bundle changes since real-origin/main", so the renderer manifest is untouched and `publisher:release:verify` is not owed here. That is the repo's own classifier answering, not my reading of the path list.

CI has not run yet at the time of writing.

## Summary by CodeRabbit

- **New Features**
  - Added consistent page messages for loading, unavailable, error, and signed-out states.
  - Added reusable loading, unavailable, and sign-in prompt components.
  - Updated asset, faction, and ruleset pages with clearer state messaging and navigation links.

- **Documentation**
  - Added Storybook examples for page messages and shared state components.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -53,6 +53,10 @@ export default defineMain({
       titlePrefix: 'Widgets/Asset Face',
     },
     {
+      directory: '../src/app/widgets/page-message',
+      titlePrefix: 'Widgets/Page Message',
+    },
+    {
       directory: '../src/app/ui/block',
       titlePrefix: 'Blocks',
     },
--- a/src/app/routes/_app/assets/$type/$slug/index.tsx
+++ b/src/app/routes/_app/assets/$type/$slug/index.tsx
@@ -16,8 +16,9 @@ import type { RulesetAssetSlot } from '@shared/rulesets/assetSlots';
 import type { ErrorComponentProps } from '@tanstack/react-router';
 import { createFileRoute, Link, notFound } from '@tanstack/react-router';
 import { LoadError } from '@ui/block/LoadError';
+import { LoadPending } from '@ui/block/LoadPending';
+import { NotAvailable } from '@ui/block/NotAvailable';
 import { OpenableTile } from '@ui/block/OpenableTile';
-import { PageTitle } from '@ui/block/PageTitle';
 import { Section } from '@ui/block/Section';
 import { formatRelativeDate } from '@ui/content/dates';
 import { ProfileLink } from '@ui/content/ProfileLink';
@@ -50,6 +51,7 @@ import { loadAssetPage, useAssetPage } from '@app/db/assets';
 import type { AssetPageData } from '@app/db/assets';
 import { isStaleClientData } from '@app/db/core/clientBoundary';
 import { AssetFace } from '@app/widgets/asset-face/AssetFace';
+import { PageMessage } from '@app/widgets/page-message/PageMessage';
 
 import { useAssetDeletion, useAssetGroupActions } from '../../-assetEditorStates';
 import { compositionTiles, DUPLICATED_TILE_CAP, omissionNote } from './-composition';
@@ -76,37 +78,31 @@ export const Route = createFileRoute('/_app/assets/$type/$slug/')({
 });
 
 /**
- * The frame this page wears before it has an asset: loading, absent, failed to load.
- * One component because all three are the same page with different words;
- * the three detail pages before this each repeat the markup two or three times.
+ * The words `PageMessage` wears on this route: the asset type's own name, and a way back to that type's browse page.
+ * A local component rather than a repeated prop because the type is a route param and every message on this page reads it.
+ * An unknown type still gets a frame, since the loader's `notFound` throw has to land somewhere.
  */
 function AssetDetailMessage({ children }: { children: ReactNode }) {
   const { type } = Route.useParams();
   const label = isAssetType(type) ? ASSET_TYPES[type].label : 'Assets';
   return (
-    <PageLayout>
-      <PageLayout.Header>
-        <Stack align="center" gap="xs">
-          <PageTitle title={label} />
-          <Anchor renderRoot={(rootProps) => <Link {...rootProps} to="/assets/$type" params={{ type }} />}>
-            Back to {label.toLowerCase()}
-          </Anchor>
-        </Stack>
-      </PageLayout.Header>
-      <PageLayout.Content>{children}</PageLayout.Content>
-    </PageLayout>
+    <PageMessage
+      title={label}
+      back={
+        <PageMessage.Back to="/assets/$type" params={{ type }}>
+          Back to {label.toLowerCase()}
+        </PageMessage.Back>
+      }
+    >
+      {children}
+    </PageMessage>
   );
 }
 
 function AssetDetailPending() {
   return (
     <AssetDetailMessage>
-      <Surface padding="xl">
-        <Stack gap="xs">
-          <Title order={2}>Loading</Title>
-          <Text c="dimmed">This asset is still loading.</Text>
-        </Stack>
-      </Surface>
+      <LoadPending title="Loading">This asset is still loading.</LoadPending>
     </AssetDetailMessage>
   );
 }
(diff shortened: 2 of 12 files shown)


ndelangen:
Orchestrator review at 7fde2205f64d1d3ee72b575399c19aa248f6d054: **code concur — merge-ready once the pixels land**. All CI green on this head.

Verified in the diff, beyond the two calls I concurred with pre-push (the pane on the frame; three bodies not four):

- No double pane anywhere: the bodies' own `Surface padding="xl"` wrappers are gone from all three routes, every `LoadError` now sits inside a `PageMessage` frame, and no hand-rolled message frame survives on any of the three.
- The faction route's verbatim inline pending repeat (#631's finding) now collapses into one component.
- `AssetDetailMessage` is the membrane test passed: it supplies only per-type words while the frame does the rest, and its rewritten JSDoc says exactly that.
- `PageMessage.Back` matches `Links.Item`'s createLink-over-forwardRef idiom precisely.
- Stories: five, pure args, no decorators, every body represented plus the no-back variant. The `.storybook/main.ts` title-root addition follows the existing pattern.

Two notes:

1. **Process, not defect:** `.storybook/main.ts` was in the diff but not in the hand-off's file list. Benign here, but the reviewer's enumerate-the-commit-list rule has a reporter-side twin: name every file you touched.
2. **Optional:** the JSDoc carries hand-counted numerals ('about seventeen sites', 'nine of these sentences across seven routes', 'five times across four routes'). They will rot as adoption proceeds and the motivation reads fine without them. Take or leave.

Outstanding for merge-ready: the embedded before/after imagery per the standing proof rule — the release-assets flow from my message, including the LoadError floating-vs-on-pane pair and one mobile-width shot. This review lapses if the head moves past the SHA above; if the only new commit is a JSDoc numeral trim I will re-anchor on the diff read alone.

ndelangen:
## The visible change, in pictures

Method, since it decides what these are worth: every "before" was captured from a detached `real-origin/main` checkout served by the dev server, not from main's file dropped into this branch's tree. Same server, same port, same hosted data, same viewport. The only forced part is the state itself, which is why the filenames say `staged`: the error rows need the ruleset loader to throw, and the pending row needs it slowed plus a client-side navigation, because a cold load blocks in the loader before React mounts anything.

### The error body, at 1280

Before, floating on the page background:

Error frame before, desktop

After, on the pane the frame paints:

Error frame after, desktop

### The same body at 375, which is the row that changed my mind

Before. The alert runs to the content edge with no side gutter:

Error frame before, mobile

After. It keeps the pane's 32px inset, the same one the pending and absent bodies always had:

Error frame after, mobile

| | holder of the alert | painted | width at 1280 | width at 375 | inset |
|---|---|---|---|---|---|
| before | `main` | no | 1200px | 351px | 0px |
| after | the frame's `Surface` | yes | 1134px | 285px | 33px |

### And the state that should not have moved

The pending frame, before and after. These are meant to be indistinguishable, and the point of including them is that the change is confined to the error body rather than to the frame:

Pending frame before

Pending frame after

What the pictures cannot show is the one thing that did change here: the after body is inside a `status` live region, so a screen reader now hears that the page is loading. Five routes wrote this heading and sentence by hand and none of them announced it.

One measurement I am deliberately not tabulating for this pair. My probe reports "is the body's parent painted", and the parent is a different element on each side (main's `Stack` inside its own `Surface`, this branch's body directly inside the frame's `Surface`), so the two numbers are not comparable and would read as a difference that is not there. The screenshots are the honest comparison for this row.

ndelangen:
Pixels verified: heads unchanged since my code concurrence at 7fde2205f64d1d3ee72b575399c19aa248f6d054, all six images render from the release, and I read the desktop error pair myself — the alert moves from a full-bleed floating strip onto the frame's inset pane exactly as decided, with band, title, and way-back identical on both sides. The staged captions are correct per the archive convention, and pulling the paned-before/paned-after probe row rather than publishing a measurement whose subject changed between sides was the right call, documented honestly. **Ready for Norbert's merge.**

PR #727, opened 2026-08-24
The turning dice stay behind the header instead of drawing across it

ndelangen:
Spotted by Norbert while reviewing #726: the turning dice draw across the header artwork.

## Why it happens

The dice are `html::after`, fixed to the viewport at `z-index: -1`. The band artwork is also `z-index: -1`, and `.band` is `position: relative` with `z-index: auto`, so it opens no stacking context of its own. Both therefore land in the root stacking context at the same level, where paint order falls to tree order, and generated content on `html` comes last. The dice win.

`z-index: -2` settles it without touching anything else. The dice are a texture on the page background, which is what -2 means here, and the blend now composites against that background alone rather than against whatever `-1` content happened to paint first.

## When it shows

Not on every window, which is why it survived this long. The band's height follows the content column rather than the viewport (`100cqw * 1125/3064`, and the column caps at 1200px), so the band bottom settles at about 478px on any wide window. The dice sit at 60vh to 120vh. They meet once the viewport is shorter than roughly 800px.

Reproduced at 1660x700 on `/rulesets`. At 1280x860 the page looks fine on both sides of this change, so a taller window is not evidence of anything.

## Verified

Same server, same page, same viewport, only `page.css` differing between the rows.

| | dice `z-index` | band artwork `z-index` | dice over the artwork |
|---|---|---|---|
| before | -1 | -1 | yes |
| after | -2 | -1 | no |

Checked in both schemes and at both band heights: `/rulesets` light, `/rulesets` dark, and `/rulesets/create` for the compact band. The dice still show below the band in all six, so this pushes them behind the header rather than turning them off.

Gates: lint, format:check, check:prose, check:css-orphans. No TypeScript and no stories touch this file, and `page.css` is not among the stylesheets Storybook's preview loads.

One line of CSS plus its reason. Separate from #726 because it has nothing to do with the page-message widget; it was simply visible in the same window.

### DIFF
--- a/src/app/styles/page.css
+++ b/src/app/styles/page.css
@@ -17,7 +17,11 @@ html[data-mantine-color-scheme="dark"] {
 html::after {
   content: "";
   position: fixed;
-  z-index: -1;
+  /* One step further back than the page's own content, which sits at -1: the band artwork and the
+     dice shared that level, so tree order decided, and generated content on `html` paints last.
+     The dice therefore drew across the header on any window short enough for the two to meet. At
+     -2 it stays what it is, a texture on the page background, and blends against that alone. */
+  z-index: -2;
   bottom: -40vh;
   left: 0;
   width: 100%;


ndelangen:
Orchestrator review at b5a78874ad574c27a2414a9639631669212622e0: **code concur — one requirement before merge-ready: pixels**.

The stacking analysis is verified: at -2 the dice stay above the root background (negative stacking levels still paint over the root element's own background) and drop below everything at -1, and the overlay blend now composites against the photograph alone. The six-cell verification table (schemes × band heights, plus the dice-still-visible check) and the ~800px reproduction constraint make this reviewable without trusting anyone's memory, and the 'a taller window is not evidence of anything' line deserves to be quoted in future visual passes.

The one gap is the standing proof rule: a visual-surface PR carries embedded imagery. You now have the mechanism (release-assets flow, per my message): one before/after pair at 1660x700 on /rulesets — before from a main-serving server showing the dice across the header, after from this branch — uploaded as pr-727-*.png and embedded in a comment. Then this is ready for Norbert.

ndelangen:
## The overlap, in pictures

Both captured at 1660x700 on `/rulesets`, which is the window shape where the two layers meet. "Before" is a detached `real-origin/main` checkout served by the dev server; "after" is this branch on the same server, same port, same hosted data. Nothing is staged: this is the page as it loads.

Before. The dice arc cuts up into the sand, across the lower third of the artwork:

Catalogue before, dark

After. The arc is clipped at the band's edge and the artwork is clean, while the dice keep their place below it:

Catalogue after, dark

The same pair in the light scheme, since the dice blend with `mix-blend-mode: overlay` and the two schemes give that blend different backdrops:

Catalogue before, light

Catalogue after, light

Worth repeating from the description, because it is what makes this reviewable: at 1280x860 both sides look correct and there is nothing to see. The band's height follows the content column's width and caps near 478px, while the dice are pinned to 60vh, so the two only meet once the viewport is shorter than roughly 800px. A screenshot at a comfortable desktop size proves nothing here either way.

ndelangen:
Pixels verified: head unchanged since my concurrence at b5a78874ad574c27a2414a9639631669212622e0, all four images render, and the light pair proves the claim beyond argument — the wireframe draws across the band artwork and title before, and sits behind the page after, with the texture still present on the background (pushed behind, not turned off, as the six-cell table said). One capture-timing note for the record: the ruleset card's artwork differs between the two shots; that is image load timing, not this fix, and it is not part of the evidence. **Ready for Norbert's merge.**

ndelangen:
The four images above have been replaced in place, so the URLs are unchanged and the comparison is now controlled. Worth explaining, because the reason turned out to be more interesting than the timing guess.

The ruleset thumbnail that differed between the two shots is **hosted off-site**, at `i.postimg.cc`. It is not the app's asset and not the dev server's, so whether it appears in a screenshot is a fact about someone else's CDN. Measured across four runs: it decoded in six seconds once, never in thirty on the next, and on one run it arrived for one page of a pair and not the other. That last case is the dangerous one, because it produces exactly the mismatched pair seen here while every wait in the script reports success.

So it is no longer left to chance. Both sides now stub that one external request with a transparent pixel, which is why both tiles are empty rather than one empty and one loaded. Stating it plainly since a substituted pixel that goes unmentioned is the thing that makes proof shots worthless: the stub applies identically to before and after, it is the only request intercepted, and nothing in the app's own rendering is touched.

Aborting the request instead was the first attempt and looked worse, since a broken-image glyph reads as a defect of its own. Every image on the page now reports `complete` with a non-zero natural width at shutter time on both sides, which the capture asserts rather than assumes.

The pair is now identical except for the arc across the band, which is the change under test.

ndelangen:
Re-verified after the in-place replacement: I re-read the light pair from the same URLs. The controlled pair is now the sole-difference ideal — both sides identical including the stubbed cover, the arc across the band the only change — and the disclosed transparent-pixel stub is the right way to pin a third-party variable. The earlier capture-timing footnote is obsolete and this pair replaces it as the evidence. **Ready for Norbert's merge**, unchanged: head still at b5a78874ad574c27a2414a9639631669212622e0.

Issue #729, opened 2026-08-25
Every page has a story showing its main variations


ndelangen:
## Question

Stated by Norbert, 2026-08-25: every route/page should have a Storybook story showing that page in its main variations. This ticket carries the execution once the way is settled.

Known at charting time:

- Precedent exists at both ends: the AppRoot stories mount full pages through `ShellStoryPage.stories.fixture` (which scopes `page.css` to the story's lifetime), and the page-message stories render whole page frames with real bodies.
- For detail pages the variations are now largely enumerable, and the page-message widget made most of them cheap: loaded (with real-shaped data), pending, failed, not found, and signed-out are one frame with different bodies.
- The mounting mechanism is settled by [Build and merge the browser-local Convex Storybook foundation](https://github.com/ndelangen/dunezone/issues/742): a story mounts the actual route component through a memory router and supplies deterministic database, identity, path, and search state outside the page.
- Composition refactors are not a blocker. Because stories render the unchanged route rather than copying its markup, they can provide deterministic visual and interaction coverage while [Work the composition wave](https://github.com/ndelangen/dunezone/issues/701) changes the page internals. The bounded variation contract is settled in [Define main variations for each page kind](https://github.com/ndelangen/dunezone/issues/746).

Acceptance: every visible page outcome under `src/app/routes` is represented in Storybook in its useful main variations; route-only exclusions are accounted for in the pull request, and the Storybook browser test run exercises the stories.

## Consequence Norbert attached, 2026-08-25

Once these stories exist, **Storybook becomes the main target for proof images**: mock data makes a before/after pair stable against the dev deployment's changing rows, which is the variable that already forced one proof pair to be re-shot with a stubbed request. Two requirements this adds to the mechanism decision:

- Stories must be deterministic to the pixel: fixed fixture data, no wall-clock or random values in what renders, external assets stubbed or vendored.
- The scheme axis needs a story-level answer. Storybook's iframe forces its own colour scheme today (emulation measured inert), which is why both-scheme captures currently require the real dev server; if stories are to carry proof, a globals-driven scheme switch that stamps the app's own attribute is part of the mechanism.

The real dev server remains the source for the varied-real-data browser pass on new visual features: mock data is controlled, and controlled data is also conveniently unrepresentative, which is the trap that let two defects through every gate once before.

ndelangen:
The composition-wave dependency has been removed.

[...]

PR #730, opened 2026-08-25
The four editor pages say why you cannot edit, in the frame everything else uses

ndelangen:
Second slice of item 1 on [#701](https://github.com/ndelangen/dunezone/issues/701), on the shape merged in #726: the pane belongs to the frame, and the bodies are blocks beside `LoadError`.

Eleven guard frames across the four editor routes. The remaining five sites, all one-line login gates on create pages, follow in the last slice.

## What changed on each route

| route | frames | becomes |
|---|---|---|
| `rulesets/$rulesetSlug/edit` | absent ×2, profile loading, signed out, denied | four frames, since the two absent branches were identical and are now one condition |
| `factions/$factionId/edit` | signed out, loading, denied | three frames |
| `groups/$groupSlug/edit` | absent, signed out, denied | three frames |
| `profiles/$profileSlug/edit` | signed out, not yours | two frames |

Every denied state gains a heading it did not have, so a screen reader meets an outline entry rather than a loose sentence. The two profile guards gain a page title, having previously rendered no `h1` at all.

**One way back per frame, replacing a toolbar of two icon-only buttons.** Where a guard carried a toolbar, its back arrow and its view-the-entity icon become a single named link, and the destination chosen is the entity rather than the step backwards: a reader who cannot edit a ruleset can usually still read it. That is a real trade, two affordances for one, and the one that remains is a named link rather than an unlabelled glyph.

**The identity band stays #451's.** Where a guard wore the band, the frame now carries the same name as its title. A message page says what it is about in words; the band with its cover art belongs to the page that has something to edit.

## A live defect on main, fixed here

`/groups/<bad-slug>/edit` and `/factions/<bad-slug>/edit` render a raw stack trace today.

Both loaders throw when the slug names nothing, rather than returning a null the component could check, so the component's own absent branch never runs. Neither route defined an `errorComponent`, so TanStack's unstyled default caught it and printed the Convex error into the page.

Verified on main rather than inferred: check out `e6b4fca925a`, run the dev server, and open either URL.

Each route now answers with the same frame everything else uses and the `LoadError` body, which is the case #719 added that block for. The reader still sees the server's own sentence, exactly as on the five routes #700 covered; what they no longer see is a stack trace under a broken header.

This rides with the slice rather than travelling separately because it is the same shape adoption in the same two files. Happy to cherry-pick it out if you would rather review it alone.

## Something worth knowing that this PR only half-fixes

On main, the ruleset editor's identity band renders its cover image **overflowing the band**, spilling outside the header and squeezing the title into a three-line wrap. It is visible in the before shot below at 1280x860.

The guard frames no longer show that band, so this slice removes it from the states it touches. The loaded editor page composes the same band and I could not reach it, since it needs a signed-in owner. So: this is not fixed, it is merely no longer on the pages I converted, and it belongs to #451 along with the rest of that band. Flagging it rather than letting the before/after imply a fix I did not make.

## Proof

Anonymous, against the worktree dev server. Every row is a URL a signed-out reader can type; nothing is staged, which is why no filename says so. Each "before" is a detached `real-origin/main` checkout on the same server and port, so only the checkout differs.

| state | route | `h1` before → after | way back before → after |
|---|---|---|---|
| signed out | `/rulesets/dreamrules/edit` | Edit Dreamrules → Edit Dreamrules | two icon buttons → "Back to ruleset" |
| absent | `/rulesets/nope/edit` | Edit ruleset → Edit ruleset, plus an `h2` | two icon buttons → "Back to rulesets" |
| signed out | `/factions/spacing-guild/edit` | Edit Spacing Guild → unchanged | "Back to faction" → unchanged |
| absent | `/factions/nope/edit` | **none, stack trace** → Edit faction | none → "Back to factions" |
| signed out | `/groups/testgroup/edit` | Edit testgroup → unchanged | "Back to group" → unchanged |
| absent | `/groups/nope/edit` | **none, stack trace** → Edit group | none → "Back to profiles" |
| signed out | `/profiles/central/edit` | **none at all** → Profile settings | none → "Back to profiles" |

Every frame sits on the frame's pane, and none overflows its viewport at 1280x860.

**Not covered, and worth stating plainly.** The three denied frames need a signed-in reader who is not the owner, and the two pending frames need a slow answer from the profile query. I reached neither, so five of the eleven frames are verified by types and gates only. They are the frames whose bodies (`NotAvailable`, `LoadPending`) are already exercised by the states above on other routes.

Images follow in a comment.

## Verified

Local gates: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 732 unit tests, 375 storybook tests. `bun run check` reports no capture-bundle changes, so the renderer manifest is untouched.

No e2e spec asserts on any copy this changes; I grepped the ten specs for the changed sentences and confirmed the grep was against real files rather than an empty directory.

### DIFF
--- a/src/app/routes/_app/factions/$factionId/edit.tsx
+++ b/src/app/routes/_app/factions/$factionId/edit.tsx
@@ -1,13 +1,14 @@
-import { Alert, Anchor, Stack, Text } from '@mantine/core';
+import { Alert, Stack, Text } from '@mantine/core';
 import { isRouteNoticeCode } from '@shared/routeNotices';
 import type { RouteNoticeCode } from '@shared/routeNotices';
-import { createFileRoute, Link, useNavigate } from '@tanstack/react-router';
-import { PageTitle } from '@ui/block/PageTitle';
+import { createFileRoute, useNavigate } from '@tanstack/react-router';
+import { LoadPending } from '@ui/block/LoadPending';
+import { LoginGate } from '@ui/block/LoginGate';
+import { NotAvailable } from '@ui/block/NotAvailable';
 import { factionAuthoringStatusMessage } from '@ui/content/assetPublishingStatus';
 import { ConfirmDeleteAction } from '@ui/control/ConfirmDeleteAction';
 import { IconAction } from '@ui/control/IconAction';
 import { PageLayout } from '@ui/layout/PageLayout';
-import { Surface } from '@ui/surface';
 import { UserRoundMinus } from 'lucide-react';
 import { useRef, useState } from 'react';
 
@@ -23,6 +24,7 @@ import type { FactionAuthoringViewHandle } from '@app/widgets/faction-editor/Fac
 import { FactionGroupPopover } from '@app/widgets/faction-editor/FactionGroupPopover';
 import { FactionLoadPopover } from '@app/widgets/faction-editor/FactionLoadPopover';
 import { useFactionAuthoring } from '@app/widgets/faction-editor/useFactionAuthoring';
+import { PageMessage } from '@app/widgets/page-message/PageMessage';
 
 import { useFactionNameField } from '../-factionNameField';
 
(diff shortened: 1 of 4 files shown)


ndelangen:
Four pairs, at 1280x860, anonymous. Each "before" is a detached `real-origin/main` checkout on the same dev server and port, so the checkout is the only difference. Nothing is staged: every one of these is a URL a signed-out reader can type.

## The stack trace, which is the part that is a defect rather than a refactor

`/groups/there-is-no-such-group/edit` on main:

Group edit, bad slug, before

And on this branch:

Group edit, bad slug, after

`/factions/there-is-no-such-faction/edit` is the same story, and the same two lines fix it:

Faction edit, bad slug, before

Faction edit, bad slug, after

The sentence in the after is still the server's own, verbose and full of request ids. That is `LoadError` behaving as specified and as it does on the five routes #700 covered; what changed is that it is now inside a page rather than instead of one.

## The band, and the thing this PR does not fix

`/rulesets/dreamrules/edit` signed out, on main. The cover image is **outside the band**, over the nav and past the header's lower edge, and the title is wrapped to three lines around it:

Ruleset edit, signed out, before

On this branch the guard frame carries the ruleset's name as a page title instead:

Ruleset edit, signed out, after

Read that pair carefully, because the honest description is narrower than it looks. **The band is not fixed.** It is no longer rendered on the states this slice converts. The loaded editor page composes the same band from the same code, and I could not reach it, since it needs a signed-in owner. The overflow belongs to #451 with the rest of that band, and I would rather say so than let a clean "after" imply I repaired it.

This pair is also the one where the cover art matters, so unlike the #727 shots nothing is stubbed here: both sides waited for every image to decode and asserted it before the shutter.

## A page that had no name

`/profiles/central/edit` signed out. Before, the page rendered no `h1` at all and offered nothing but the sentence:

Profile edit, signed out, before

After:

Profile edit, signed out, after

## A note on how to read the "way back" column

My capture probe matched link **text**, so it reported the ruleset editor's guards as having no way back. They do have one on main: a toolbar with two icon-only buttons, a back arrow and a view-the-ruleset glyph, both visible in the before shot above. The body's table says "two icon buttons" rather than "none" for those rows because I checked the pixels before writing it, and the rows that do say "none" are the three where the page genuinely offered nothing: the two stack traces, which render no page at all, and the profile guard, which you can see above has no header, no `h1` and no link but the one in its sentence.

Worth stating because the direction differs by row. Two unlabelled glyphs becoming one named link is a trade with the count going down. Nothing becoming a named link is not.

ndelangen:
Orchestrator review at 3c4dbd05e009b8e3804b8d6a7e33546780fbbf1e (both commits enumerated: 5d36c7b8ebd extraction, 3c4dbd05e00 defect fix): **concur, ready for Norbert's merge** once `audit` finishes (everything else green as I write).

Verified in the diff: all eleven guard frames replace their hand-rolled predecessors with none left over; the rulesets route's two identical not-found blocks collapse into one condition, and its band and toolbar are now built only after every guard has returned, with the band explicitly left to #451; the factions guard order keeps the conditional title with its reasoning stated in place; both new errorComponents carry accurate constraint comments. I read the group-edit pair personally: the raw stack trace on main is proven and the fix is proven, and the ride-along travels exactly as agreed — own section, repro, before shot.

Adjudications:

1. **The five browser-unverified frames are accepted as disclosed.** The three denied states need a signed-in non-owner, and per house rule the signed-in glance is Norbert's alone — those three join his one-sitting queue. The two pending states render the same LoadPending the verified routes exercise; argument-not-sighting is honestly labelled, and #729's stories will make all five capturable later.
2. **Guard copy is normalized, not preserved** — titles added, sentences aligned to the shape's voice (e.g. 'Loading profile…' becomes a titled 'Loading your profile'). Slice 1 preserved copy verbatim; this slice does not, visibly and I believe correctly, but Norbert should know the difference when he compares pairs.
3. **Out of scope, one line:** the LoadError body prints the server's full message including request id and `convex/groups.ts:56:25`. That is the #700-accepted shape, so nothing here — noted only in case error-message hygiene ever becomes its own ticket.

PR #731, opened 2026-08-25
The five remaining login gates say it the same way

ndelangen:
Third slice of item 1 on [#701](https://github.com/ndelangen/dunezone/issues/701), and the last of the login gates. Five sites, one shape.

With this, all nine gates the route sweep counted use `LoginGate`, and the assets feature's own shared frame was already there.

## The drift, measured

Four of the five rendered the login link as the browser's default blue rather than the themed anchor. That is not a judgement about which looks better; it is the same offer looking like two different things depending on which page you reached.

| gate | link colour before | after |
|---|---|---|
| `/factions/create` | `rgb(248, 175, 64)` | unchanged |
| `/groups/create` | `rgb(158, 158, 255)` | `rgb(248, 175, 64)` |
| `/rulesets/create` | `rgb(158, 158, 255)` | `rgb(248, 175, 64)` |
| `/rulesets/$slug/faq/create` | `rgb(158, 158, 255)` | `rgb(248, 175, 64)` |
| `/admin/migrations` | `rgb(158, 158, 255)` | `rgb(248, 175, 64)` |

Read against `LoginGate`'s own JSDoc, which claims three of nine sites were themed and six were not: those three are the faction create and the two editor gates that landed in #726 and #730, so the claim holds across the whole set rather than just this slice.

## Three per-site calls

**`/rulesets/create` becomes an early return.** Its gate was a ternary inside the content, so the page kept its toolbar and swapped the form for a sentence. A page that cannot be used is a different page, not this one with a hole in it, and the toolbar's lone back arrow is what the frame's way back replaces.

**`/admin/migrations` keeps having no way back.** The route has no parent in the navigation and today's version offers none, so the frame omits it rather than inventing a destination. Where an admin page should send a signed-out reader is a question about admin navigation, and I would rather leave it visible than answer it here. It is the one row in the table below with `ABSENT` on both sides.

**The FAQ create page gets only its two message frames.** Its loaded state still renders a hand-rolled `h1` inside `PageLayout.Header`, which is item 5 of this wave rather than item 1. So for now that page spells the same words two ways depending on whether you are signed in. Flagging it because it is a mid-state I am knowingly leaving, not one I missed.

## Proof

Signed out, at 1280x860, against the worktree dev server. Nothing is staged: all five are URLs anyone can type while logged out, which is the point of the slice. Each "before" is a detached `real-origin/main` checkout at `b5e246bdfe0` on the same server and port.

| gate | `h1` | sentence | way back |
|---|---|---|---|
| `/factions/create` | Create faction | Log in to create a faction. | Back to factions |
| `/groups/create` | Start group | Log in to start a group. | Back to profiles |
| `/rulesets/create` | Create ruleset | Log in to create a ruleset. | Back to rulesets |
| `/rulesets/dreamrules/faq/create` | Ask a question | Log in to ask a question. | Back to ruleset |
| `/admin/migrations` | Migration activity | Log in to view migration activity. | absent, as above |

Every frame sits on the frame's pane and none overflows its viewport. All five states are verified in a browser; unlike #730 there is nothing here I could not reach.

Images follow in a comment.

## Verified

Local gates: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 732 unit tests, 375 storybook tests. `bun run check` reports no capture-bundle changes.

Rebased onto `b5e246bdfe0` after #730 merged, and the gates were re-run on the rebased tree rather than the pre-rebase one.

## Summary by CodeRabbit

* **UX Improvements**
  * Standardized signed-out states across migrations, faction creation, group creation, and ruleset pages.
  * Added consistent login prompts and back navigation using shared page messaging.
  * Improved FAQ creation feedback with dedicated loading and authentication states.
  * Simplified ruleset creation access flow for signed-out visitors.

### DIFF
--- a/src/app/routes/_app/admin/migrations.tsx
+++ b/src/app/routes/_app/admin/migrations.tsx
@@ -1,5 +1,6 @@
 import { Group, Stack } from '@mantine/core';
-import { createFileRoute, Link } from '@tanstack/react-router';
+import { createFileRoute } from '@tanstack/react-router';
+import { LoginGate } from '@ui/block/LoginGate';
 import { PageTitle } from '@ui/block/PageTitle';
 import { IconAction } from '@ui/control/IconAction';
 import { PageLayout } from '@ui/layout/PageLayout';
@@ -8,6 +9,7 @@ import { RefreshCw } from 'lucide-react';
 
 import { loadAdminMigrationDashboard, useAdminMigrationDashboard, useSyncMigrationRuns } from '@db/migrations';
 import { useCurrentProfile } from '@db/profiles';
+import { PageMessage } from '@app/widgets/page-message/PageMessage';
 
 export const Route = createFileRoute('/_app/admin/migrations')({
   loader: async () => ({ dashboard: await loadAdminMigrationDashboard() }),
@@ -30,18 +32,14 @@ function AdminMigrationsPage() {
   const dashboard = dashboardQuery.data;
   const syncRuns = useSyncMigrationRuns();
 
+  /* The one gate with no way back, because this route has no parent in the navigation and today's
+     version offers none either. Where an admin page sends a signed-out reader is a question about
+     admin navigation rather than about this frame, so it is left as it was rather than invented. */
   if (!profile.data?._id) {
     return (
-      <PageLayout>
-        <PageLayout.Header>{migrationsHeader}</PageLayout.Header>
-        <PageLayout.Content>
-          <Surface padding="lg">
-            <p>
-              <Link to="/auth/login">Log in</Link> to view migration activity.
-            </p>
-          </Surface>
-        </PageLayout.Content>
-      </PageLayout>
+      <PageMessage title="Migration activity">
+        <LoginGate action="view migration activity" />
+      </PageMessage>
     );
   }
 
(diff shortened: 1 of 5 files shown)


ndelangen:
Orchestrator review at 714ed8ed66e7035b9a1107e61f000f0e11450195 (single commit, branch at rest, rebased onto #730's merge with gates re-run on the rebased tree): **concur — Norbert gets the ready-for-merge call when e2e_docker and storybook finish**, everything else green as I write.

Verified in the diff: all five gates convert faithfully with copy preserved, back destinations preserved (factions, profiles, ruleset, rulesets), and each judgment reasoned in place. The three per-site calls are all right: the rulesets-create early return is the shape's own logic (a page that cannot be used is a different page, and the toolbar goes with it); migrations keeping no way back is correct since inventing admin navigation is not a framing decision, and the widget's contract explicitly allows omitting it; the FAQ-create two-spellings tension is the honest intermediate state and its comment points at item 5. The FAQ pending state gaining a real LoadPending is a bonus beyond the sweep's count. I read the migrations pair personally: default-blue underlined link before, themed accent after, frame otherwise identical — the measured 4-of-5 drift table matches what the pixels show.

The measurement-over-memory note (commit message said three, probe said four, amended before push) is the discipline working; logged on my side too.

Issue #732, opened 2026-08-25
The live-query result claims an error channel it never uses


ndelangen:
## Question

`toLiveQueryResult` in `src/app/db/core/live.ts` hardcodes `isError: false, error: null` on every live query result (lines 39-40; the real error tracking at lines 96-97 belongs to the mutation path). The type therefore advertises an error channel that never fires, and every `if (someQuery.isError)` branch written against a live query is unreachable — which is how five routes crashed to the router's raw error overlay while each believed it had an error frame: the branch that was supposed to catch the failure could not run, and nothing else was watching. Found during the composition wave's slice 3; the two dead branches in the tree (groups detail, groups edit) are deleted there, and the five crashing routes get errorComponents (#730 and the slice 3 PR).

The decision this ticket resolves: **wire it or strike it from the type.**

- **Wire it:** make `toLiveQueryResult` report real query errors. A real data-layer change with more callers than the route tree shows; would also create two competing error postures per route (isError branch vs errorComponent), which wants a stated rule.
- **Strike it:** remove `isError`/`error` from the live-query result type entirely, so the dead branch cannot be written again, and state the contract: a live query that fails throws to the route's catch boundary, and the errorComponent is the one place a route answers it. This matches the shape the wave has been installing and makes the illegal state unrepresentable.

Ordering: decide after the composition wave's slice 3 merges (it deletes the two dead consumers, so the strike option becomes a pure type narrowing). Mutations keep their live error channel either way — theirs is real.

ndelangen:
Implementation chooses strike: live queries expose data and loading state only, failures go to the route errorComponent, and mutations keep their real error channel. [Live queries leave failures to route boundaries](https://github.com/ndelangen/dunezone/pull/735) is ready for Norbert to merge. Both required review axes passed. The hosted development database was not used.

ndelangen:
Resolved by choosing strike. [Live queries leave failures to route boundaries (#735)](https://github.com/ndelangen/dunezone/pull/735) merged at 1c1fa101f20ad78537fa1621a1f6f8af632c411b. LiveQueryResult now exposes only data and loading state. Failed queries go to the route errorComponent, while mutations keep their real error channel. The pull request contains before-and-after visual proof of the contract change.

PR #733, opened 2026-08-25
The last four message frames join the rest, and three more crashes land on a page

ndelangen:
Last slice of item 1 on [#701](https://github.com/ndelangen/dunezone/issues/701). Groups detail, profiles detail, the faction catalogue and the FAQ question page.

With this, every site the route sweep counted under finding 1 wears the shared frame, and finding 3's nine login gates all use `LoginGate`.

## A class defect, five routes, two causes, one fix

Three more routes render the router's error overlay for a slug that names nothing:

- `/groups/there-is-no-such-group`
- `/profiles/there-is-no-such-profile`
- `/rulesets/dreamrules/faq/there-is-no-such-question`

With the editor pair fixed in #730, that is five routes in this wave. Verified on main at `b5e246bdfe0` rather than inferred: open any of those three and you get a stack trace where a page should be.

Two causes. Groups detail and profiles detail throw from their loaders, like the editor pair. The FAQ route's loader already catches and returns `notFound`; what throws is its live query, after mount. An `errorComponent` answers both, because a route's catch boundary covers render as well as load.

**A guard that could not fire.** The FAQ page's not-found frame was written inside the component body, below a hook that calls `useQuery` unconditionally. A missing question therefore threw during the same render that would have shown the frame, so a question that does not exist read as a question that failed to load. The guard now sits in a thin route component above the subscription, and the body it used to share moved into `LoadedFaqQuestion`. `AssetDetailPage` splits for exactly this reason and says so in its own comment.

**A branch that could never run.** Groups detail carried its own `groupData.isError` frame. `toLiveQueryResult` reports `isError: false` for every query, so that branch was unreachable, which is why the page crashed instead of showing it. It is deleted rather than converted. The seam itself is [#732](https://github.com/ndelangen/dunezone/issues/732), which holds the wire-it-or-strike-it decision; deleting this consumer and the half-dead one in groups edit is what makes the strike option a pure type narrowing.

## Two deliberate losses

**The groups detail skeleton grid.** Four grey boxes in a two-column layout, the only skeletons in the tree, replaced by the same "Loading group" every other page shows. Skeletons preserve layout and read as fast; this one also announced nothing to a reader who cannot see it, and was a second spelling of a state the wave exists to spell once. If you would rather keep skeletons, the right move is a kit block for them and adoption everywhere, not one page keeping its own.

**The faction catalogue's header on its two frames.** They lose the eyebrow, the description and the create-a-faction call to action, since the frame carries a title and a way back and nothing else. On the pending frame that is a transient loss; on the error frame the call to action was arguably wrong anyway. This is the one place in three slices where the shared frame costs a reader something they could have used, and I would rather you weighed it than discovered it.

## Proof

Signed out, at 1280x860, against the worktree dev server. Every row is a typeable URL. Each "before" is a detached `real-origin/main` checkout at `b5e246bdfe0` on the same server and port.

| route | before | after |
|---|---|---|
| `/groups/nope` | stack trace, no `h1`, no way out | `h1` Group, alert body, Back to profiles |
| `/profiles/nope` | stack trace, no `h1`, no way out | `h1` Profile, alert body, Back to profiles |
| `/rulesets/dreamrules/faq/nope` | stack trace, no `h1`, no way out | `h1` FAQ, `h2` Question not found, Back to ruleset |
| `/factions` | loads normally | unchanged |

The FAQ row is the one worth reading twice. Before the guard moved it showed the alert body, meaning it had reached the `errorComponent`; after the split it shows "Question not found" with no live region, which is the not-found body doing its job. Both states are correct pages, but only the second tells the reader the truth.

Images follow in a comment.

## On the commit shape

One commit rather than a frames commit and a fixes commit. Two of these files carry both concerns, and the error component introduces the back-link constant the converted frames use, so neither half compiles alone. I would rather say that than offer a split that does not build.

## Verified

Local gates: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 732 unit tests, 375 storybook tests. `bun run check` reports no capture-bundle changes.

### DIFF
--- a/src/app/routes/_app/factions/index.tsx
+++ b/src/app/routes/_app/factions/index.tsx
@@ -4,7 +4,6 @@ import {
   Drawer,
   Group,
   InputBase,
-  Loader,
   Popover,
   RangeSlider,
   Select,
@@ -17,6 +16,7 @@ import { Link, createFileRoute } from '@tanstack/react-router';
 import type { ErrorComponentProps } from '@tanstack/react-router';
 import { FactionCatalogueSpotlight } from '@ui/block/FactionCatalogueSpotlight';
 import { LoadError } from '@ui/block/LoadError';
+import { LoadPending } from '@ui/block/LoadPending';
 import { PageTitle } from '@ui/block/PageTitle';
 import { complexityTierSliderMarks } from '@ui/content/ComplexityGlyph';
 import { formatFactionCatalogueDate } from '@ui/content/dates';
@@ -33,6 +33,7 @@ import type { KeyboardEvent } from 'react';
 import { loadFactionCataloguePage, useFactionCataloguePage } from '@db/factions';
 import type { FactionCatalogueEntry, FactionCataloguePageData, FactionRulesetSummary } from '@db/factions';
 import { isStaleClientData } from '@app/db/core/clientBoundary';
+import { PageMessage } from '@app/widgets/page-message/PageMessage';
 
 import {
   complexityRangeSearchValue,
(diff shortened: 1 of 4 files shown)


ndelangen:
The images this PR's body promised, posted after the merge rather than before it. My session was killed between opening the PR and uploading them, so the promise went unkept while the change went in. Recording that plainly, because a body that says "images follow in a comment" and then has no comment is the kind of small untruth that makes the rest of a proof section less believable.

They were captured before the merge, at 1280x860, signed out, each "before" from a detached `real-origin/main` checkout at `b5e246bdfe0` on the same dev server and port. They are worth keeping despite arriving late: the before state no longer exists on main and reproducing it now means checking out an old base.

## The crash

`/groups/there-is-no-such-group` on main before this PR. A stack trace where a page should be, with the Convex request id and the handler line visible to anyone who types a wrong URL:

Group detail, bad slug, before

After. The route's `errorComponent` catches it, and the reader gets a named page and a way out:

Group detail, bad slug, after

## The guard that could not fire

`/rulesets/dreamrules/faq/there-is-no-such-question`, same before state:

FAQ question, bad slug, before

After. This is the row worth reading twice. An `errorComponent` alone would have shown the red alert body, because the live query throws during the same render that would have shown the not-found frame. Moving the guard above the subscription is what produces this instead, and the difference between the two is whether the reader is told the question does not exist or told that something failed:

FAQ question, bad slug, after

## One thing this PR set up that has since happened

The body said the seam cause was [#732](https://github.com/ndelangen/dunezone/issues/732)'s to decide, and that deleting the two dead `isError` consumers would make the strike option a pure type narrowing. That is what was chosen: `LiveQueryResult` on current main declares `data`, `isPending` and `isLoading` only, with `isError` and `error` kept on `LiveMutationResult` where they are real. The dead branch is now unwritable rather than merely deleted.

PR #735, opened 2026-08-25
Live queries leave failures to route boundaries

### LINKED ISSUE #732: The live-query result claims an error channel it never uses

ndelangen:
## What changed

- Remove the `isError` and `error` fields that every live query returned as `false` and `null`.
- Remove the remaining dead group-edit check.
- Document one query-failure path: Convex throws to the route's `errorComponent`.
- Leave mutation error state unchanged.

The prerequisite composition slice is already on main through #733.

## Visual proof

Both images use the same ready query input. The base commit returns five fields, including an error channel that can never fire. This branch returns three fields and makes `query.isError` a TypeScript error.

| Before at `618c02a581b` | After at `b27760b40e3` |
|---|---|
| Before: the live query result includes isError false and error null | After: the live query result contains only data, isPending, and isLoading |

This change deliberately leaves the rendered route error frame alone. Failed Convex queries still reach the route `errorComponent`. The change removes the false result-level alternative.

## Verification

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run check:prose`
- `bun run check:app-layout`
- `bun run check:css-orphans`
- `bun run test` (728 tests)
- `VITE_CONVEX_URL=https://exuberant-finch-263.eu-west-1.convex.cloud bun run publisher:release:verify`

The shared hosted development database was not used.

Closes #732

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified live query results to expose only loading and data states.
  * Updated group pages to display “Group not found” appropriately when group data is unavailable.
  * Failed subscriptions now use the route’s error component for error handling.

* **Documentation**
  * Updated state-management guidance to reflect the revised live query and error-handling behavior.

### DIFF
--- a/docs/state-management.md
+++ b/docs/state-management.md
@@ -39,9 +39,10 @@ export function useFaction(slug: string, options?: { initialData?: FactionDetail
 ```
 
 `toLiveQueryResult` exists to keep call sites uniform: it returns
-`{ data, isPending, isLoading, isError, error }`, falling back to `initialData` while the
-subscription is still undefined. `isError` is always `false`, because a failed Convex subscription
-throws to the route's error boundary rather than resolving into a result object.
+`{ data, isPending, isLoading }`, falling back to `initialData` while the subscription is still
+undefined. A failed Convex subscription throws to the route's `errorComponent` rather than
+resolving into a result object. The route handles the failure there instead of carrying a second,
+unreachable error branch in its component.
 
 **Example**: [`src/app/db/factions.ts`](../src/app/db/factions.ts)
 
--- a/src/app/db/core/live.ts
+++ b/src/app/db/core/live.ts
@@ -12,8 +12,6 @@ export type LiveQueryResult<TData> = {
   data: TData | undefined;
   isPending: boolean;
   isLoading: boolean;
-  isError: boolean;
-  error: Error | null;
 };
 
 export type LiveMutationResult<TVariables, TResult> = {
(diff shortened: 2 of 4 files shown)


ndelangen:
Orchestrator review at b27760b40e30704c1653048a66da884ec08266ce: ready for Norbert to merge. Standards: no findings. Spec: no findings. The review lapses if the head moves.

Issue #752, opened 2026-08-25
Adopt formatted text in every stored-prose editor and reader


ndelangen:
## Question

Adopt formatted text in every textarea-like stored-prose editor and app reader. Wire `FormattedTextInput` into faction rule bodies, treachery-card text, Asset About, ruleset description, FAQ questions, and FAQ answers; render parsed blocks through the app Content renderer wherever the app displays those values. Re-run the production classifier before each field family narrows, and make shared Zod write parsing the authority.

Use field profiles instead of one permissive shape everywhere:

- Setup, revival, and FAQ questions are marks-only inline fields. They may use bold, italic, and underline, but may not create paragraphs, lists, or hard breaks.
- Prose bodies may use the full formatted-text block grammar.
- Single-line identity labels and SVG token labels remain plain text.
- Legacy Rulebook code remains out of scope; the Rulebooks map owns its replacement.

The inventory and measured collision set are recorded in [Inventory the prose fields and measure the markdown collision](https://github.com/ndelangen/dunezone/issues/670). The parser, editing control, app Content renderer, and game renderers are already shipped.

ndelangen:
Delivered through PR #753 and PR #754. Production now rejects paragraphs and line breaks in setup, revival, and FAQ-question fields while retaining bold, italic, and underline marks. Prose bodies use the full block grammar, all writes pass through shared Zod schemas, and app readers render parsed formatted text. Deployment run 32863867052 passed the exact Worker smoke test and the production narrowing check. The before/after production proof is embedded in PR #754.

PR #753, opened 2026-08-25
Faction inline prose is ready for formatted text

ndelangen:
Part of #752.

This prerequisite release adds the shared marks-only formatted-text profile and normalizes the one production faction value that still contains layout line breaks. The narrow editor and reader changes follow after production completes this migration.

## Production evidence before migration

Richese currently renders without a visual layout regression. The migration changes stored structure, not the published sheet appearance. I will post the production after image and pixel comparison after deployment.

Richese before migration

## Checks

- bun run test
- bun run typecheck
- bun run migrations:static-check
- VITE_CONVEX_URL set to production: bun run check
- VITE_CONVEX_URL set to production: bun run publisher:release:verify
- Shared-development migration and verification completed successfully

### DIFF
--- a/convex/migration-guards.json
+++ b/convex/migration-guards.json
@@ -197,6 +197,16 @@
       "requires": ["faction_complexity_grouped_v1"]
     },
     {
+      "id": "faction_inline_formatted_text_v1",
+      "phase": "widen",
+      "requires": []
+    },
+    {
+      "id": "faction_inline_formatted_text_verify_v1",
+      "phase": "widen",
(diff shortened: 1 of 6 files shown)


ndelangen:
## Production before and after

The production migration and its verification both completed successfully across all 38 faction rows. Richese now stores its setup as one inline value, and the published sheet remains pixel-identical: 0 changed pixels on page 1.

| Before | After |
| --- | --- |
| Richese before migration | Richese after migration |

Production release: https://github.com/ndelangen/dunezone/actions/runs/32861501647

Issue #756, opened 2026-08-25
Prove a faithful browser Convex runtime or stop the map


ndelangen:
## Question

Can Storybook execute every Convex query and mutation path required by application pages against deterministic browser-local state without corrupting execution context or maintaining a misleading partial runtime?

The earlier browser prototype proved real page execution, then the context-conformance work found that Convex global helpers and service objects depend on ambient execution state. This ticket starts from that blocker. It does not repeat the demo or erase the failed conformance evidence.

## Work

- Inventory the Convex functions and components reachable from current application pages.
- Test current releases, prereleases, upstream work, and the smallest credible pinned fork or adapter.
- Preserve one isolated worker per active story, serialized calls within a story, deterministic reset, and parallel execution across stories.
- Exercise representative real pages through their production queries and page-owned mutations.
- Cover authentication, nested calls, component paths, triggers, rollback, scheduling, query refresh after mutation, story switching, and repeated reset where page-reachable code uses them.
- Run the same conformance suite in Storybook development and a served static build below a non-root path.
- Keep the worker offline and use only synthetic state. Do not claim or contact the hosted development database.
- Measure implementation size, upstream coupling, unsupported behavior, maintenance cost, and failure clarity.
- Fail explicitly when a page reaches unsupported Convex behavior. Do not replace a real handler with a response fixture.

## Evidence

- Retain every failing context case as a reproducible test.
- Show a real page before and after a query or mutation and after reset.
- Store the visual receipts in the shared sibling screenshot collection and link them from the pull request or prototype report.
- Record one durable report that distinguishes proven fidelity from limitations.

## Done when

One of two outcomes is established:

1. A faithful and maintainable browser runtime passes the page-reachable conformance matrix in development and static builds, giving the implementation ticket a precise foundation to merge.
2. The remaining blocker requires an unacceptable private runtime, produces incorrect behavior, or cannot fail safely. The map stops without falling back to mocked query responses.

ndelangen:
## Checkpoint: the Zone-backed runtime passes the hard browser cases

I tested the current releases, `convex@1.45.0` and `convex-test@0.0.56`. Neither release adds browser context support. Upstream issue get-convex/convex-test#122 still identifies native `await` as the reason a Promise shim cannot implement `AsyncLocalStorage`, and PR get-convex/convex-test#83 adds more ambient isolation rather than a browser carrier.

A new prototype makes the ambient contract work without editing Convex or `convex-test`:

[...]

Issue #742, opened 2026-08-25
Build and merge the browser-local Convex Storybook foundation


ndelangen:
## Outcome

Build and merge the browser-local Convex page-story foundation accepted by the preceding research, prototype, and decision tickets.

The runtime gate in #756 passed. This implementation must execute real Convex handlers and must not replace them with response fixtures.

## Scope

- Execute the repository's real Convex query and mutation handlers against deterministic browser-local state.
- Provide one Storybook-only adapter at the app's Convex doorway while preserving each page's production query and mutation code.
- Start every page story from one canonical valid database. The story may mutate that baseline or return a valid replacement, including an empty database.
- Derive table types from the Convex schema and validate domain content through existing shared authorities. Deterministic helpers fill mechanical values; meaningful page content stays explicit in the story.
- Keep identity, route parameters, and search parameters separate from the database parameter.
- Run actual routes, page composition, forms, page-owned mutations, navigation intents, and mutation-driven query refresh.
- Give each active story an isolated worker and restore fresh declared state before every run.
- Fail clearly when a page reaches unsupported Convex behavior. Do not return a plausible mocked result.
- Prove authentication, component paths, nested calls, triggers, rollback, scheduling, isolation, story switching, and reset for every behavior reachable from application pages.
- Prove Storybook development, the static build, a served non-root base path, and the existing Storybook browser CI.
- Document page-story authoring, validation failures, runtime limitations, and the boundary between stories, direct tests, and journeys.
- Keep the hosted development database out of the workflow.
- Run the repository's required release verification before opening or updating the pull request.

The public story API is:

```tsx
export const Populated = meta.story({
  parameters: {
    database: db((baseline) => {
      baseline.factions.push(faction({ name: 'House Atreides' }));
    }),
  },
});
```

## Public Storybook

- Publish a separate, secret-free static deployment at https://storybook.dune.zone; keep the application at https://dune.zone.
- Bundled repository Convex source is acceptable because the repository is public. Production credentials, deployment URLs, production data, and live `/published/*` dependencies remain forbidden.
- Use synthetic deterministic data only.
- Copy the canonical `public/` directory into the Storybook artifact so existing root-relative images, vectors, fonts, and textures remain same-origin. Do not create a second maintained asset source.
- Enforce the accepted CSP on Storybook documents and worker responses.
- Add `bun run verify:storybook-publication` as a mandatory CI gate over final static bytes, served headers, and browser network behavior.
[...]

Issue #739, opened 2026-08-25
Deterministic Storybook page scenarios


ndelangen:
## Destination

Prove whether Storybook can faithfully execute this repository's real Convex page queries and page-owned mutations against deterministic browser-local state in development and static builds. If the runtime is maintainable, merge it as the reusable page-story foundation and then give every application page stories for its main variations. If it is not, stop this map and report the blocker. Do not fall back to mocked query responses.

## Notes

- This map carries execution. It closes only when the browser-local Convex foundation is merged and tested and the separate page phase is complete, or when the runtime hard gate proves the plan infeasible.
- The runtime hard gate and foundation are complete. The page phase begins after its bounded variation contract is settled and may proceed alongside route-composition work.
- A page scenario combines deterministic database state, path parameters, and search parameters, then renders one real route through its production query and mutation code.
- Stories represent meaningfully distinct page versions, not a generated parameter or error matrix.
- Stories use compact synthetic state and never claim or depend on the single hosted development database.
- Every story is isolated, rerunnable, and exercised through the existing Storybook browser CI in development and static delivery modes.
- End-to-end tests remain unchanged during this map. A later coverage audit will decide which simple real-stack journeys remain.
- Every meaningful visual change needs before-and-after proof stored in the shared sibling screenshot collection and shown in the pull request.
- Research uses primary sources and records one durable Markdown report per ticket.
- Application changes receive a standards and specification review before handoff.
- This map is a child of [The UI audit: taxonomy, shared patterns, data needs, honest docs](https://github.com/ndelangen/dunezone/issues/628). Page stories mount unchanged routes, so they may support and proceed alongside [Work the composition wave](https://github.com/ndelangen/dunezone/issues/701).

## Decisions so far

- [Can Convex run inside a Storybook Web Worker?](https://github.com/ndelangen/dunezone/issues/736) - Development and static Storybook can run real page queries and authenticated mutations, but the first browser context shim corrupts component paths after overlapping or component-heavy execution.
- [Define the deterministic Storybook seed contract](https://github.com/ndelangen/dunezone/issues/743) - Research found that fixed-time actors, domain recipes, aliases, generated mechanical values, and real page mutations can replace broad fixture dumps; the later authoring decision sets the final baseline and validity rules.
- [Set the security boundary for browser-hosted Convex source](https://github.com/ndelangen/dunezone/issues/741) - The artifact review found no secret values, hosted Convex URL, or production data and established the build, CSP, data, and network controls.
[...]

PR #759, opened 2026-08-25
The browser Convex prototype passes its hard gate

ndelangen:
## Verdict

Proceed to the foundation phase in #742.

This evidence-only prototype runs the real Rulesets create page against browser-local Convex handlers inside a Storybook Web Worker. The page executes its production mutation, navigates to its production detail route, and reads the committed row through `detailPageBySlug`. It works in Storybook development and a static build served below `/catalogue/`, including while the browser is offline.

The workaround is deliberately fail-closed: Zone.js supplies the worker execution context, Oxc lowers the explicit worker dependency closure to Promise continuations, and AST plus emitted-bundle checks reject retained native async syntax. No hosted Convex deployment or shared development database was contacted.

## Visual proof

The first pair is the hard-gate before and after. The retained failure used the browser `AsyncLocalStorage` shim without complete async lowering; three of five checkpoints resumed in the wrong frame. The current build preserves every frame and completes the Convex global helper call.

| Before: ambient context is wrong | After: context and helper pass |
| --- | --- |
| Three of five execution-context checkpoints use the wrong frame | All execution-context checks and createFunctionHandle pass |

The second sequence shows the production page seam rather than story-only output. The create route writes through the production mutation, the production detail route reads the result, and worker replacement returns to a clean create page.

| Before mutation | After mutation and production query | After worker replacement |
| --- | --- | --- |
| The real Create ruleset page before mutation | The real Ruleset detail page after mutation | The real Create ruleset page after worker replacement |

The full report is in [`prototype-receipts/756/browser-convex-runtime.md`](https://github.com/ndelangen/dunezone/blob/norbert/prototype-756-browser-convex-runtime/prototype-receipts/756/browser-convex-runtime.md).

## Verification

- 749 unit tests passed.
- 394 of 394 browser stories passed.
- All four hard-gate stories passed against the static build below `/catalogue/`.
- The static page story replaced its worker and created `OfflineCreatedRuleset` while the browser was offline.
- Typecheck, lint, prose, formatting, Storybook build, artifact scan, and `publisher:release:verify` passed.
- Final specification and repository-standards reviews found no remaining issues.

## Limits

- Dependency upgrades can change the worker closure, so the transform allowlist and conformance gates must be maintained.
- Query refresh after mutations is coarse; this does not reproduce Convex dependency-tracked subscriptions.
- Each active story owns a worker and browser-local database.
- Compact deterministic seed data is still required; the runtime removes response fixtures, not domain state.
- Full Auth provider flows are not proven. Current identity-based page handlers are.

Supports #756 and map #739. This PR is evidence only and is not intended to merge.


ndelangen:
Closing this completed evidence-only spike. The PR was never intended to merge. Its accepted result shipped through #762, with publication follow-ups #765, #766, and #767, and #742 closed after live verification. The prototype, report, and visual proof remain available here for reference.

Issue #797, opened 2026-08-27
The stragglers the audit found still hand-roll their frames


ndelangen:
## Question

The item-1 audit's remaining drift, batched: `profiles/$profileSlug/delete.tsx` hand-rolls four whole-page frames with no header band (the page loses its own name in all four states, and the way back appears and disappears by state); the homepage at `_app/index.tsx` is half-converted (LoadError bare in Content with no Surface at ~:420-433; a hand-rolled pending frame whose bare Surface resolves to padding none at ~:401-418); `AppNotFound` and `[_]_jobs.tsx` still hand-roll and were outside the original sweep's scope.

Convert each to PageMessage plus the right body, or record why a site stays (AppNotFound may argue it IS the frame). One PR, before/after pairs from the page stories.

ndelangen:
Closing the loop on the staged homepage stories deferred from here to #769: they are consciously not built. PageMessage.stories.tsx now exports the exact frames (Loading, Failed, NoWayBack), so a homepage staging would prove the component's contract a second time through a route that owns nothing in the exchange; the residue is shared route config read at review time, with router.tsx's defaultErrorComponent as the designed fallback. Reasoning in full on PR #817.

Issue #796, opened 2026-08-27
The asset editor family speaks the message vocabulary


ndelangen:
## Question

Surfaced by the item-1 audit on [Work the composition wave](https://github.com/ndelangen/dunezone/issues/701): `AssetEditorMessage` in `routes/_app/assets/-assetEditorStates.tsx` is a second, independently built page-message frame — PageLayout, compact header, PageTitle, Surface padding xl, its own back anchor — with twenty-two render sites across eleven files plus five DriftedAssetPage and two NoEditorYet routing through it. It predates the merged PageMessage widget, so it is uncovered ground rather than regression, but its bodies are bare Text: the login gates at -bundleCreate.tsx and -bundleEdit.tsx hand-roll the sentence LoginGate exists for, and its pending states carry no status live region.

The task: the asset editor family adopts PageMessage and the four bodies, retiring AssetEditorMessage. The asset detail page already speaks the vocabulary; this brings its edit and create siblings along. Largest remaining population from the audit.

PR #800, opened 2026-08-27
Asking a question about a ruleset that is not there lands on a page

ndelangen:
A miss of mine, found by the item-1 audit on [#701](https://github.com/ndelangen/dunezone/issues/701#issuecomment-5444066429) and fixed here.

`rulesets/$rulesetSlug/faq/create` has the same defect the five `errorComponent` fixes in #730 and #733 addressed. It sits in the same directory as `faq/$questionSlug.tsx`, which slice 3 fixed, and it already imports `PageMessage` and already renders `PageMessage` frames for its pending and signed-out states. I converted its frames and did not notice its loader.

Not in #796 or #797, which cover the asset editor family and the remaining hand-rolled frames. This is a one-route bug fix and it is mine, so it goes on its own rather than riding with someone else's ticket.

## The defect

`loader: async ({ params }) => ({ ruleset: await loadRulesetBySlug(params.rulesetSlug) })`, unguarded, and the query throws rather than returning nothing:

- `src/app/db/rulesets.ts:96-99` calls `api.rulesets.getBySlug`
- `convex/rulesets.ts:94-98` delegates to `loadRulesetPublicBundleBySlug`
- `convex/lib/rulesetDetailPage.ts:51-55` throws `Ruleset with slug ${slug} not found`

With no route-level `errorComponent`, that lands on the router's default. Worth stating precisely, because it is easy to say wrongly: the route is **not** unguarded. `src/app/router.tsx:28` sets `defaultErrorComponent: AppErrorComponent`, which is styled for the stale-client case and otherwise returns TanStack's `ErrorComponent`. So the reader gets the raw block rather than nothing at all, and the fix upgrades a default into a page rather than installing a missing boundary.

## One deliberate difference from its sibling

`faq/$questionSlug.tsx`'s error frame sends the reader back to the ruleset. This one sends them to the catalogue. The distinction is which thing is missing: on the question page the ruleset exists and only the question is absent, so "back to ruleset" leads somewhere real; here the ruleset is the thing that is not there, and the same link would offer the reader the page that just failed.

## Proof, and why the story is not a rubber stamp

The proof is a page story rather than a screenshot pair: `Rulesets/AskQuestionMissingRuleset`, at `/rulesets/there-is-no-such-ruleset/faq/create`. It mounts the real route against the seeded database through the same clone seam every page story uses, so it runs in `storybook:test` in CI from now on.

A story that merely renders proves nothing here, because `AppErrorComponent` renders without throwing too. So I checked that the story actually discriminates, by removing the error component and rendering the same story again with everything else identical:

| | alert body | way back | router's "Something went wrong!" |
|---|---|---|---|
| without `errorComponent` | none | absent | present |
| with it | "This ruleset could not be loaded" | Back to rulesets | absent |

That is the whole claim: the same story, the same seed, the same path, and the only difference is the four lines this PR adds.

## Known and out of scope

The rendered message is the raw `error.message`, which in a dev build carries a long Convex stack. That is the error-message-internals observation already recorded as out of scope during the #730 review, and it applies to every `LoadError` caller rather than to this route. I have not touched it, but the story now makes it visible to whoever picks it up.

## Verified

Local gates on a fresh worktree at `8c2c37d187e`: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 767 unit tests, 445 storybook tests (444 before; the new story is the extra one). `bun run check` reports no capture-bundle changes.

### DIFF
--- a/src/app/routes/-rulesets.stories.tsx
+++ b/src/app/routes/-rulesets.stories.tsx
@@ -13,6 +13,13 @@ export const Edit = meta.story({ args: { path: '/rulesets/classicrules/edit' } }
 export const AskQuestion = meta.story({
   args: { path: '/rulesets/classicrules/faq/create' },
 });
+/**
+ * The same page reached through a ruleset slug that names nothing, which is the state the route's own frame exists for.
+ * Its loader throws rather than returning nothing, so without that frame this path falls to the router's default and renders the error unstyled.
+ */
+export const AskQuestionMissingRuleset = meta.story({
+  args: { path: '/rulesets/there-is-no-such-ruleset/faq/create' },
+});
 export const Question = meta.story({
   args: { path: '/rulesets/classicrules/faq/when-does-the-storm-move' },
 });
--- a/src/app/routes/_app/rulesets/$rulesetSlug/faq/create.tsx
+++ b/src/app/routes/_app/rulesets/$rulesetSlug/faq/create.tsx
@@ -1,6 +1,8 @@
 import { Button, Group, Stack } from '@mantine/core';
 import type { FaqTag } from '@shared/faq/tags';
+import type { ErrorComponentProps } from '@tanstack/react-router';
 import { createFileRoute, Link, useNavigate } from '@tanstack/react-router';
+import { LoadError } from '@ui/block/LoadError';
 import { LoadPending } from '@ui/block/LoadPending';
 import { LoginGate } from '@ui/block/LoginGate';
 import { RulesetLink } from '@ui/content/RulesetLink';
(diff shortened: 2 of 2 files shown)


ndelangen:
Orchestrator review at 2a1fda4ae00bf381c63e418177795b8db99e2c17 (single commit, branch at rest, based on the chips merge): **concur — ready call follows when e2e_docker and storybook finish**, everything else green.

The fix is exactly the sibling pattern with the one deliberate difference reasoned in place: the way back points at the catalogue because on THIS route the ruleset is the thing that is missing, where the question page's ruleset exists and can be returned to. Owning the miss rather than folding it into #797 was right — it was a gap in delivered work, not uncovered ground, and a four-line PR with its own proof is cheaper to trust than a line item in someone else's batch.

The discrimination table is the part that outlives this PR: a story offered as proof of an error frame must show it discriminates — same story, same seed, same path, the guard removed — or it proves only that SOMETHING rendered. That is the vacuous-guard rule applied to stories, and it enters the review protocol as of now: any story cited as proof of an error or empty state carries its without-the-guard counterpart in the PR body. Adopted, and going in my memory.

PR #802, opened 2026-08-27
The asset editors speak the message vocabulary instead of a second one

### LINKED ISSUE #796: The asset editor family speaks the message vocabulary

ndelangen:
Closes [#796](https://github.com/ndelangen/dunezone/issues/796), the largest population the item-1 audit found.

`AssetEditorMessage` was a second page-message frame, built independently and predating the widget: its own `PageLayout`, its own compact header, its own `Surface`, its own anchor below the words. Twenty-two render sites across eleven files, plus `DriftedAssetPage` and `NoEditorYet` routing through it, and every body a bare `Text`.

## What moved and what stayed

**The frame is `PageMessage`'s.** What stays in `-assetEditorStates.tsx` is the part that is genuinely this feature's: a known asset type goes back to its own browse page and anything else to the landing. That is a fact about the registry rather than about message frames, and `AssetDetailMessage` on the detail route has been exactly this shape since #726, so the two siblings now differ only in which browse page they name.

I kept the name rather than deleting it, which is a deviation from the ticket's wording ("retiring AssetEditorMessage"). The duplicate *frame* is what is retired; the type-to-destination logic would otherwise be spelled at twenty-two call sites. If you would rather see the identifier go, it is a rename plus twenty-two touches and I will do it.

**The bodies became the four blocks.** Ten hand-rolled login sentences are `LoginGate`, five not-found and five denied states are `NotAvailable`, and the way back moved from inside the pane to the band.

**Two states needed a title split.** The frame title was carrying the state rather than the page: an editor whose asset is missing announced itself as "Deck not found" and had no page name at all. It now says "Edit deck" in the band with "Deck not found" as the body's heading, which is the same split `AssetDetailMessage` uses. `NoEditorYet` reads the same way, with the type's own label in the band.

**The drifted-asset state stays bespoke**, and deliberately: it ends in an action rather than in words, so its heading and sentence are a `NotAvailable` and the offer to delete sits beside it in the same slot. Its message stays the caller's, because that JSDoc claim turned out to still be load-bearing for exactly one caller: four editors say "the token schema" or "the deck schema" after their own noun, and the treachery editor says "the treachery card schema" after "card". I checked before collapsing it, and did not collapse it.

## Proof, with the discrimination the protocol now asks for

Three new page stories, each with `play` assertions rather than a bare mount, following the `SignedOut` precedent in `rulesets/-create.stories.tsx`.

| story | asserts |
|---|---|
| `Assets/EditDeckSignedOut` | the "Log in" link, the way back, **and that the way back is not inside `main`** |
| `Assets/EditDeckMissing` | "Edit deck" in the band **and** "Deck not found" as the body heading |
| `Assets/CreateStormCardNoEditor` | "Storm cards" in the band **and** "No editor yet" as the body heading |

And the A/B, run by reverting every converted file to `real-origin/main` and leaving the three stories in place:

| | with the conversion | against unconverted main |
|---|---|---|
| `EditDeckSignedOut` | pass | **fail** |
| `EditDeckMissing` | pass | **fail** |
| `CreateStormCardNoEditor` | pass | **fail** |

The signed-out story is the one that needed thought. Its words do not change: that route already used a themed anchor, so nothing visible differs except where the way back sits. Asserting the position is what makes it discriminate instead of passing against either version; without that line it was a story that would have gone green on main.

Signed out on a deck editor

A slug naming no deck

A type with no editor yet

## Two things found on the way, neither fixed here

**The create pages' signed-out state is not story-reachable.** I tried `CreateDeckSignedOut` first and it rendered the full deck editor, Save button and all, to an anonymous viewer. That is not an app bug: `useCurrentProfile` documents at `src/app/db/profiles.ts:93` that `session.profile` is `null` for a signed-out viewer, and the create gates read `profile.data === null`, which fires correctly in production. In Storybook, `identity: null` leaves `profiles.session` unresolved rather than answering "signed out", so a strict `=== null` gate never fires. Routes that gate on `!profile.data?.x` show the gate in both cases, which is why the ruleset create story works.

Two consequences worth someone's attention, neither mine to take here: the story seam cannot currently express "signed out" for anything reading the profile session, and the app has two gate spellings that differ during the pending window, so an anonymous visitor briefly sees the deck editor and never the ruleset one. I used the edit route instead, whose gate reads `viewerAccess.viewer.kind` and is decided server-side.

**One storybook run failed under load.** Running `storybook:test` concurrently with the unit tests and typecheck, `EditDeckSignedOut` timed out once. Three subsequent runs alone passed 448/448. My three stories each mount a full route with 30s waits, matching the existing precedent's timeout, so this is contention rather than a defect, but three more heavy page stories is a real addition to that suite's cost and I would rather say so than have CI discover it.

## Verified

Rebased onto `2c13e30996c` and re-run there, not on the pre-rebase tree. Local gates: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 767 unit tests, 448 storybook tests. `bun run check` reports no capture-bundle changes.

### DIFF
--- a/src/app/routes/-assets.stories.tsx
+++ b/src/app/routes/-assets.stories.tsx
@@ -1,4 +1,5 @@
 import preview from '@sb/preview';
+import { expect, within } from 'storybook/test';
 
 import { pageStoryMeta } from './-storybookConfig';
 
@@ -42,4 +43,46 @@ export const EditEnhanceToken = meta.story({
 });
 export const EditBundle = meta.story({
   args: { path: '/assets/bundle/atreides-tokens/edit' },
+});
+
+/**
+ * The editor reached by a reader who is not signed in.
+ * The assertions are the point rather than the render: this state and the two below all rendered before this frame existed, so a story that only mounts them would go green either way.
+ *
+ * The edit route rather than the create route, because this one asks the server: its gate reads `viewerAccess.viewer.kind`, which the asset query decides from the identity, while the create pages read the profile session, which the story seam leaves unresolved rather than answering "signed out".
+ */
+export const EditDeckSignedOut = meta.story({
+  args: { path: '/assets/deck/house-treachery/edit' },
+  parameters: { identity: null },
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+    await expect(page.findByRole('link', { name: 'Log in' }, { timeout: 30_000 })).resolves.toBeVisible();
+    const back = await page.findByRole('link', { name: 'Back to decks' }, { timeout: 30_000 });
+    await expect(back).toBeVisible();
+    /* The words here did not change, only who owns them: this route's own frame used to put the way
+       back inside the pane, below the sentence, and the shared frame puts it in the band. Asserting
+       the position is what makes this story discriminate rather than pass against either version. */
+    expect(back.closest('main')).toBeNull();
+  },
+});
+
+/** A slug that names no deck. The page keeps its own name in the band and says what is missing in the body. */
+export const EditDeckMissing = meta.story({
+  args: { path: '/assets/deck/there-is-no-such-deck/edit' },
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+    await expect(page.findByRole('heading', { name: 'Edit deck' }, { timeout: 30_000 })).resolves.toBeVisible();
+    await expect(page.findByRole('heading', { name: 'Deck not found' }, { timeout: 30_000 })).resolves.toBeVisible();
+    await expect(page.findByRole('link', { name: 'Back to decks' }, { timeout: 30_000 })).resolves.toBeVisible();
+  },
+});
+
+/** A type the registry knows and nothing can author yet, which is a statement about the roadmap rather than a missing page. */
+export const CreateStormCardNoEditor = meta.story({
+  args: { path: '/assets/card-storm/create' },
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+    await expect(page.findByRole('heading', { name: 'Storm cards' }, { timeout: 30_000 })).resolves.toBeVisible();
+    await expect(page.findByRole('heading', { name: 'No editor yet' }, { timeout: 30_000 })).resolves.toBeVisible();
+  },
 });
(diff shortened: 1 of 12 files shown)


ndelangen:
Orchestrator review at db934a9bfdbb5b3c9d8c5f42dbb41b0b32f9f3e0 (single commit, branch at rest, rebased and re-gated on the rebased tree): **concur — ready call follows CI**, which was mid-run as I reviewed.

**The deviation is accepted as flagged, no rename wanted.** What the ticket meant by retiring AssetEditorMessage was the duplicate frame, and that is retired; what remains under the name is the type-to-destination registry knowledge, which is exactly the membrane AssetDetailMessage already models, and its rewritten JSDoc says so. A 22-site rename would buy a name at the cost of churn and lose the symmetry with its sibling.

**The proof section sets the new standard.** Three play-asserted stories, the A/B run by actually reverting every converted file rather than by reasoning, and the signed-out story's position assertion is the discrimination rule earning its keep in the same PR that adopted it — a story that would have passed on main, caught and given the line that fails there. The checked-before-not-collapsing note on the drifted-asset state is the read-the-comment-before-classifying discipline in action.

Spot-read verified: the bundle and deck login gates carry their copy into LoginGate action phrases faithfully; the reshaped AssetEditorMessage delegates the frame and keeps only what is genuinely the asset registry's.

Both unfixed findings are dispositioned on the map: the story seam's signed-out gap goes to the behavioural-validation ticket on the Storybook map, and the two-gate-spellings divergence gets its own small ticket. This review lapses if the head moves past the SHA above.

ndelangen:
SonarCloud's red is adjudicated, not ignored: the quality gate fails on `new_duplicated_lines_density` alone (6.8% against a 3% threshold), with zero issues raised and reliability, security, and maintainability all OK. The 'duplicated lines' are the twenty-two now-uniform message frames — the same short `AssetEditorMessage` + body blocks at every site — which is the uniformity this PR exists to install, not duplication to extract: the shared code already lives in the components, and what repeats at the call sites is the caller's words. An advisory metric misreading convergence as copy-paste does not outrank the required gates, which are green.

Issue #769, opened 2026-08-25
Add selective behavioural validation to page stories


ndelangen:
## Outcome

After the route stories exist and page composition has settled, add selective behavioural coverage where a page's own contract benefits from it.

## Scope

- Review the stories delivered by [Every page has a story showing its main variations](https://github.com/ndelangen/dunezone/issues/729).
- Add semantic assertions or `play` interactions only where they prove useful page behaviour: a primary mutation, mutation-driven query refresh, navigation, deletion, mode switching, or another route-owned action.
- Keep read-only and primarily visual stories free of assertion ceremony unless an assertion catches a realistic regression.
- Test page behaviour, not multi-page journeys. A story may exercise its page's query and related mutations without becoming an end-to-end flow.
- Keep exhaustive unhappy query branches and server invariants in direct tests.
- Do not generate an assertion matrix or require the same checks for every page kind.
- Preserve the existing end-to-end suite during this stage. The later measured coverage audit decides retirement.

## Coordination

This stage waits for both the page-story sweep and [Work the composition wave](https://github.com/ndelangen/dunezone/issues/701). Story creation can support composition work, but detailed assertions couple more closely to controls, labels, and page structure.

## Visual proof

For any visible behaviour changed while adding coverage, capture matching before-and-after receipts in the shared sibling screenshot collection and show them in the pull request.

## Done when

The useful page-owned behaviours selected during the review have focused Storybook interactions running in browser CI, deliberate omissions are documented, and no story has been expanded into an application journey.

ndelangen:
From the asset-editor conversion (#802), a seam gap this ticket inherits: the story seam cannot currently express a signed-out viewer for anything that reads the profile session — `identity: null` leaves `profiles.session` unresolved rather than answering signed-out, so strict `=== null` gates never fire in a story even though they fire in production. Confirmed against src/app/db/profiles.ts:93 before reporting. Behavioural coverage of login gates depends on the seam answering signed-out properly.

ndelangen:
Recording the coverage gap from [#817](https://github.com/ndelangen/dunezone/pull/817) here, so it survives that PR's merge as tracked work rather than as prose in a closed thread.

## Login gates that read the profile session cannot be behaviourally covered

[...]

PR #804, opened 2026-08-27
The straggler frames join the vocabulary

### LINKED ISSUE #797: The stragglers the audit found still hand-roll their frames

ndelangen:
Closes [#797](https://github.com/ndelangen/dunezone/issues/797). The four sites the item-1 audit left, converted rather than exempted.

## The account deletion page

Four hand-rolled whole-page frames and no header band in any of them, so the page lost its own name in every state and its way back appeared and disappeared depending on which one you landed in.

The interesting part is its denied branch. It read `!data || data.kind === 'denied'` and answered both with one sentence plus a login link, so a reader who was already signed in as the wrong person was invited to log in. The server has always told the two apart, sending `reason: 'signed_out'` or `reason: 'wrong_profile'` (`convex/accountDeletion.ts:139` and `:143`), and the page discarded it. They are now a `LoginGate` and a `NotAvailable`.

Its in-progress state was one `Alert` whose colour flipped to red on failure under an unchanged "in progress" title. Those are two states, and they are now `LoadPending` (`status`, polite) and `LoadError` (`alert`, assertive) with a title that says the operation did not finish.

The finished state is the one thing here with no block: a completed action is neither absent, pending nor failed, and all four bodies are ways of saying a page has nothing to show. It keeps a bare sentence, and gains the name and the way out it never had.

## The homepage

Half-converted, which is the failure mode the widget's own doc names. `HomepageError` used the body and not the frame, so its alert sat directly on the page background while every other caller's sat on a pane. `HomepagePending` used neither and hand-rolled a `Surface` with no `padding` prop, which resolves to `none` against the frame's `xl`.

## The publication jobs dashboard

Its first three states were arms of a ternary inside the content, so a reader who may not see the queue at all still got the dashboard's briefcase, its title and its description above the refusal. They are early returns now, and the header is reserved for the page that actually shows a queue.

## AppNotFound, converted rather than exempted

The ticket allows recording why a site stays, and flags that this one "may argue it IS the frame". I converted it, and the argument is that it is not a frame but a page: the most-reached message in the application, rendering a bare `<p>` and a raw `<Link>` under a title with no pane.

**This is the first place the shell installs a widget rather than only the kit,** and that deserves your eye. The dependency runs the legal direction: `.oxlintrc.json` restricts only `src/app/ui` from reaching into shell, widgets and routes, and its own message says the kit is imported by the shell, by widgets and by routes and never the reverse. The alternative was to keep the hand-rolled frame and swap only the words, which is precisely the half-conversion this same PR removes from the homepage. If you would rather the shell stayed widget-free, the fallback is to leave `AppNotFound` as it is and record it as an exemption; I did not want to pick that quietly.

Its title also changes, from "404 - Page Not Found" to "Page not found" in the band with "This page does not exist" as the body's heading. That is the same title split the asset editors needed: the frame title was carrying the state rather than naming the page.

## Proof

Two page stories, both with assertions rather than a bare mount, and both checked against a full revert.

| story | asserts | with the conversion | against unconverted main |
|---|---|---|---|
| `Site/NotFound` | "Page not found" in the band, "This page does not exist" as the body heading, and the way back **not inside `main`** | pass | **fail** |
| `Site/PublicationJobsSignedOut` | the "Log in" link, **and that the dashboard's description is absent** | pass | **fail** |

The A/B was run by reverting all five changed source files to `real-origin/main` with the stories left in place, not by reasoning about what would happen.

**Two sites are not story-covered, and I would rather say so than imply otherwise.** The homepage's pending and error frames need a staged loader or a thrown one, and I did not add stories that stage them. The account deletion page has no page story; it does have a unit suite (`-delete.test.tsx`) that covers three of its lifecycle states and the fail-closed case, which I kept passing and did not weaken.

## One thing this surfaced about the widget

`PageMessage.Back` is built with `createLink`, whose `useLinkProps` reads router context and throws outside a `RouterProvider`. The delete page's unit test mocked `Link` but not `createLink`, so converting the page broke four tests with `Cannot read properties of null (reading 'isServer')` rather than an assertion failure. The fix is one line in that file's existing router mock, and it is worth knowing generally: **any unit test that renders a `PageMessage` frame must mock `createLink` or provide a router.** The four tests then failed honestly, on the one marker my copy change had invalidated, and that marker moved with the words.

## Verified

Local gates on a fresh worktree at `2c13e30996c`: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 767 unit tests, 446 storybook tests. `bun run check` reports no capture-bundle changes.

### DIFF
--- a/src/app/routes/-site.stories.tsx
+++ b/src/app/routes/-site.stories.tsx
@@ -1,4 +1,5 @@
 import preview from '@sb/preview';
+import { expect, within } from 'storybook/test';
 
 import { pageStoryMeta } from './-storybookConfig';
 
@@ -7,8 +8,39 @@ const meta = preview.meta({
   ...pageStoryMeta,
 });
 
-export const NotFound = meta.story({ args: { path: '/a-page-that-does-not-exist' } });
+/**
+ * The most-reached message in the application, and the last one to wear the shared frame.
+ * The assertions distinguish the frame from what stood here before, which was a bare paragraph and a raw link under a title with no pane.
+ */
+export const NotFound = meta.story({
+  args: { path: '/a-page-that-does-not-exist' },
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+    await expect(page.findByRole('heading', { name: 'Page not found' }, { timeout: 30_000 })).resolves.toBeVisible();
+    await expect(
+      page.findByRole('heading', { name: 'This page does not exist' }, { timeout: 30_000 })
+    ).resolves.toBeVisible();
+    const back = await page.findByRole('link', { name: 'Go back home' }, { timeout: 30_000 });
+    expect(back.closest('main')).toBeNull();
+  },
+});
 export const Icons = meta.story({ args: { path: '/__icons' } });
 export const PublicationJobs = meta.story({ args: { path: '/__jobs' } });
+
+/** The job queue reached by a reader the server does not recognise, which is a login gate rather than an alert inside a dashboard header. */
+export const PublicationJobsSignedOut = meta.story({
+  args: { path: '/__jobs' },
+  parameters: { identity: null },
+  play: async ({ canvasElement }) => {
+    const page = within(canvasElement.ownerDocument.body);
+    await expect(page.findByRole('link', { name: 'Log in' }, { timeout: 30_000 })).resolves.toBeVisible();
+    /* The dashboard's own header, with its briefcase and its description, is what the frame replaces. */
+    await expect(
+      page.queryByText(
+        'Inspect the durable work queue and control whether the next scheduled run may pick up pending work.'
+      )
+    ).toBeNull();
+  },
+});
 export const FuturePlans = meta.story({ args: { path: '/future-plans' } });
 export const Privacy = meta.story({ args: { path: '/privacy' } });
--- a/src/app/routes/_app/[_]_jobs.tsx
+++ b/src/app/routes/_app/[_]_jobs.tsx
@@ -1,26 +1,17 @@
-import {
-  Alert,
-  Badge,
-  Center,
-  Code,
-  Group,
-  Loader,
-  Pagination,
-  Select,
-  Stack,
-  Switch,
-  Table,
-  Text,
-} from '@mantine/core';
+import { Alert, Badge, Center, Code, Group, Pagination, Select, Stack, Switch, Table, Text } from '@mantine/core';
 import { createFileRoute } from '@tanstack/react-router';
+import { LoadPending } from '@ui/block/LoadPending';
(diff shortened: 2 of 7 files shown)


ndelangen:
Orchestrator review at 044f1f8d484a58466069a12a40ff252010aa1c10 (single commit, branch at rest): **concur, with one small addition requested — then ready call follows CI.**

**The ruling you asked for: the AppNotFound conversion stands.** The shell installing the page-message widget runs the dependency in the legal direction, and more to the point, the router's notFoundComponent IS a page — the shell is standing where a route would stand, so it wears what a page wears. Your own JSDoc makes the argument I would have made, including why the alternative was the half-conversion this PR removes two files away. The exemption route is rejected.

**The deletion-page split is the find of the wave's tail**: the server always distinguished signed_out from wrong_profile and the page collapsed them into a login link shown to a signed-in reader — restoring the distinction because the vocabulary made it cheap is exactly what the vocabulary is for. The coverage table's honesty is accepted as stated: two sites story-proven with full-revert A/Bs, the homepage's staged pending/error stories deferred (they land with the behavioural ticket #769), the deletion page held by its unit suite, kept passing and unweakened.

**The one addition before ready: put the createLink fact in the widget's JSDoc** — one sentence on PageMessage saying a unit test that renders the frame must mock createLink or supply a router, since useLinkProps throws outside a RouterProvider with an error that reads as anything but its cause. That fact cost you twenty minutes and is exactly what JSDoc is for; AGENTS.md is too far from the point of use. Push it and I re-anchor on the one-line diff.

Also noted for Norbert at merge: the most-reached message's copy changes ('404 - Page Not Found' → 'Page not found' plus a fuller sentence) — a copy improvement riding the conversion, visible in the pair. This review lapses if the head moves past the SHA above.

ndelangen:
Re-anchored at dee944e48ffdc2f2a7290108279f24205e9e9f45: the delta from the reviewed tip is exactly the requested three-line testing-contract JSDoc on PageMessage, naming the symptom and not just the rule — which is the part that will save the twenty minutes. **Ready for Norbert's merge** with functional checks green. This review lapses if the head moves past the SHA above.

PR #805, opened 2026-08-27
The live-query result stops carrying a knob nothing turns

### LINKED ISSUE #798: The live-query result carries a knob nothing turns
## Question

The vestige the #732 strike left behind, found by the item-1 audit: `toLiveQueryResult`'s second parameter `enabled` defaults to true, all 22 call sites in `src/app/db` pass the literal true, and its only effect is gating isPending/isLoading — both of which reduce to `liveData === undefined`. Same shape as the struck error channel: a contract surface advertising behaviour no caller uses.

Strike it: remove the parameter, reduce the two expressions, confirm by grep that no caller passes anything but the literal. Small, mechanical, one PR.

ndelangen:
Closes [#798](https://github.com/ndelangen/dunezone/issues/798), the vestige [#732](https://github.com/ndelangen/dunezone/issues/732)'s strike left behind.

`toLiveQueryResult`'s `enabled` parameter defaulted to `true` and every call site passed the literal `true`, so the flag only ever ANDed with a constant. Its two uses, `isPending` and `isLoading`, both reduce to `liveData === undefined`.

Same shape as the error channel #732 struck: a contract surface advertising behaviour no caller uses. And in this case there was nothing for it to express even in principle, because a subscription that should not run is not disabled here, it is not mounted. That is already the contract `AGENTS.md` sets for Pickers.

## The count, and a trap in checking it

Twenty-two call sites, all passing the literal.

Worth stating because I got it wrong first: a grep for `toLiveQueryResult(` finds **twenty**, not twenty-two. Two calls in `groups.ts` pass an explicit type argument, so the identifier is followed by `<` rather than `(` and an anchored pattern skips them. The honest check is `grep -rn "toLiveQueryResult"` and then subtracting the nine imports, which gives twenty-two and matches the ticket. I mention it because "confirm by grep that no caller passes anything but the literal" is the ticket's own instruction, and the obvious grep quietly under-reports.

## What changed

- The parameter is gone and the third argument, `initialData`, becomes the second. Both call shapes were rewritten: the plain `toLiveQueryResult(x, true, () => y)` and the type-argument `toLiveQueryResult<T>(x, true, () => y)`.
- `isPending` and `isLoading` are now `liveData === undefined` directly.
- The function gains a JSDoc it did not have, saying why waiting is the only state it reports and why there is no `enabled`. That is the part that stops this coming back: the next person to want a disabled query now meets the answer at the point of use.

No behaviour changes. Every expression this touches was already evaluating to the same value.

## Verified

Local gates on a worktree at `2c13e30996c`: typecheck, lint, format:check, check:prose, check:app-layout, check:css-orphans, knip, 767 unit tests, 445 storybook tests. `bun run check` reports no capture-bundle changes.

No proof images: nothing here renders. The claim is that twenty-two expressions were already constant-folded, and the way to check it is the grep above plus a green typecheck, not a screenshot.

### DIFF
--- a/src/app/db/accountDeletion.ts
+++ b/src/app/db/accountDeletion.ts
@@ -8,7 +8,7 @@ import { api } from '../../../convex/_generated/api';
 export type ReplacementProfile = FunctionReturnType<typeof api.accountDeletion.listReplacementProfiles>['page'][number];
 
 export function useAccountDeletionPage(profileSlug: string) {
-  return toLiveQueryResult(useQuery(api.accountDeletion.page, { profileSlug }), true);
+  return toLiveQueryResult(useQuery(api.accountDeletion.page, { profileSlug }));
 }
 
 export function useReplacementProfiles(search: string) {
--- a/src/app/db/assets.ts
+++ b/src/app/db/assets.ts
@@ -15,7 +15,7 @@ export async function loadAssetCataloguePage(): Promise<AssetCataloguePageData> 
 
 export function useAssetCataloguePage(options?: { initialData?: AssetCataloguePageData }) {
   const liveData = useQuery(api.assets.cataloguePage, {});
-  return toLiveQueryResult(liveData, true, () => options?.initialData);
+  return toLiveQueryResult(liveData, () => options?.initialData);
 }
 
 export type AssetsByTypesData = FunctionReturnType<typeof api.assets.listByTypes>;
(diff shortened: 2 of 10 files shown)


ndelangen:
Orchestrator review at b3e9778b9655896261cfa20a76060c1faa6e9336 (single commit, branch at rest): **concur — ready for Norbert's merge** with functional checks green. The strike is textbook: parameter gone, both expressions reduced to what they always computed, all twenty-two call sites rewritten with no behaviour change, and the JSDoc that stops the flag returning — a subscription that should not run is not mounted, the Pickers' own contract, now stated at the point of use beside the #732 error-channel rationale.

The find worth keeping is the grep trap: `toLiveQueryResult(` returns twenty of twenty-two because two call sites pass an explicit type argument and the identifier is followed by `<`. You noticed the count disagreed with the ticket and chased it, which is the discipline; the trap joins the grep memory, since a removal sweep 'confirmed' by that pattern would miss silently where no typecheck backstops it. This review lapses if the head moves past the SHA above.

ndelangen:
Correction to the comment above, for precision: four verify jobs were still running when I wrote 'functional checks green' — the concurrence stands, but **the ready call takes effect only when this head's CI completes**. Watch is set; Norbert gets the confirmation either way.

PR #806, opened 2026-08-27
The identity band becomes a Block, and five pages wear it

### LINKED ISSUE #451: Detail-page identity headers are hand-rolled three times over

ndelangen:
Item 3 of [the composition wave](https://github.com/ndelangen/dunezone/issues/701), executing [the identity-band territory (#451)](https://github.com/ndelangen/dunezone/issues/451) on Norbert's two decisions: **the band is a Block** (`PageIdentity` in `ui/block`), and **the band's content is fixed to the light theme** — the header's scheme-pinned paper dictates every colour, so the Block sets no ink of its own and the five per-page `--color-text` overrides die.

**The Block**: media beside name, breadcrumb above (a createLink sub-component, the `PageMessage.Back` idiom), meta as children, the h1 through `PageTitle` so the level and face rules hold. It owns the row geometry (48rem), one media scale (6→4.75→3.75rem), and the title's wrap and sub-48em step. Each page keeps what is genuinely its own: the faction its circle clip and drop shadow, the asset its card-corner clip, the profile its initials placeholder, the editor its cover pane.

**Unifications riding the convergence, each a visible change flagged for veto:**
- The **ruleset editor** adopts `size="compact"` (its tall band was the anomaly) and its title takes the paper ink — the form now starts above the fold.
- The **asset band** joins the shared scale: 64rem row → 48rem, 4.5rem face → the 6rem step.
- The **profile** h1 gains the canonical title face (was a raw 1.35rem bold) and its avatar fills the shared box (72px → the 6rem step).
- All five titles go from the dark override to the pinned paper ink, in both schemes.

Net −165 lines across the five routes' markup and CSS.

## Proof

Twenty story captures: five pages × before/after × both schemes (before = the public Storybook, main's build; after = this branch; deterministic seeds, so every pair is controlled). Embedded in a comment. I read the editor and profile pairs personally: the compact adoption, the ink change, and the face treatment land exactly as decided, with everything outside the band identical.

## Gates

typecheck, lint, format, prose, css-orphans (66 stylesheets), app-layout, knip, 767 unit tests, 448 storybook tests (three new Block stories). Changed paths are outside the renderer capture graph; CI's publisher_release confirms.

Closes #451

### DIFF
--- a/src/app/routes/_app/assets/$type/$slug/index.module.css
+++ b/src/app/routes/_app/assets/$type/$slug/index.module.css
@@ -1,25 +1,9 @@
-/* The identity band follows the faction and ruleset detail pages: media in its own column, text sharing one left edge. */
-.pageHead {
-  width: min(100%, 64rem);
-  min-width: 0;
-}
-
-/* The media column: the asset's own face, small, beside the title. */
-.pageHeadMedia {
-  flex-shrink: 0;
-  width: 4.5rem;
+/* The asset's own face filling the band's media box, keeping its card-corner clip. */
+.pageHeadFace {
+  width: 100%;
+  height: 100%;
   border-radius: var(--mantine-radius-sm);
   overflow: hidden;
-}
-
-.pageHeadText {
-  min-width: 0;
-}
-
-/* The band renders data-scheme-paper, so the title takes the band's own text colour rather than the page scheme's. */
-.assetTitle {
-  overflow-wrap: anywhere;
-  color: var(--color-text, var(--mantine-color-dark-8));
 }
 
 /* The face at reading size: large enough that a card's rules text is legible, short of becoming a poster. */
--- a/src/app/routes/_app/assets/$type/$slug/index.tsx
+++ b/src/app/routes/_app/assets/$type/$slug/index.tsx
@@ -9,7 +9,7 @@
  * Its slug stays reserved so the address survives, but the page renders the same body a slug that never existed would get, matching the ruleset detail page.
  * That is deliberately not a claim that nothing was ever here, and nothing of the deleted row reaches the client to leak.
  */
-import { Alert, Anchor, Group, Stack, Text, Title } from '@mantine/core';
+import { Alert, Group, Stack, Text } from '@mantine/core';
 import { ASSET_TYPES, holdsDeckMembership, isAssetType } from '@shared/assets/types';
 import { RULESET_ASSET_SLOTS } from '@shared/rulesets/assetSlots';
 import type { RulesetAssetSlot } from '@shared/rulesets/assetSlots';
@@ -19,6 +19,7 @@ import { LoadError } from '@ui/block/LoadError';
 import { LoadPending } from '@ui/block/LoadPending';
 import { NotAvailable } from '@ui/block/NotAvailable';
 import { OpenableTile } from '@ui/block/OpenableTile';
+import { PageIdentity } from '@ui/block/PageIdentity';
 import { Section } from '@ui/block/Section';
 import { AssetLink } from '@ui/content/AssetLink';
 import { formatRelativeDate } from '@ui/content/dates';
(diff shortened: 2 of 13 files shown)


ndelangen:
The proof matrix, all from deterministic stories (before = the public Storybook, main's build; after = this branch), light pairs embedded, dark set uploaded under the same naming with `-dark-`:

**Ruleset editor** — compact adopted, paper ink, form above the fold:
edit before
edit after

**Profile detail** — canonical face and shared avatar box:
profile before
profile after

**Ruleset detail**:
ruleset before
ruleset after

**Faction detail**:
faction before
faction after

**Asset detail** — the 64rem row and 4.5rem face joining the shared scale:
asset before
asset after

Dark pairs for all five are on the release (`pr-806-band-<page>-dark-<side>.png`), archive copies in `dunezone-screenshots/pr-806/`. Nothing staged: every shot is a rendered story.

ndelangen:
Rebased onto main past the cover-rehost merge (#799). One conflict, the ruleset editor's header, resolved as the union both branches wanted: the PageIdentity band with the rehost branch's derived `coverUrl` (falling back thumb→full→legacy in the db module) instead of the raw `image_cover` the pre-rebase tree read. New tip 8e666baed579971ccc1a918a770c08ae38fdee7e; my authorship of both sides of the conflict makes this a verification rather than a re-review, and honest disclosure: the first push of this rebase briefly carried a formatter failure because a tail-pipe swallowed the gate's exit code — the same trap adjudicated on #799 tonight, now bitten by me — caught and amended within minutes, with gates re-verified on bare exit codes. All checks re-running; ready call follows CI.

ndelangen:
**Norbert's catch, fixed at 9d26d57ba51**: the asset face was cut off because the shared media box forced a 6rem square with `overflow: hidden` onto media that was never square — the old asset band constrained width only. The media column now owns width alone, and each square media declares its own `aspect-ratio: 1` (faction symbol, profile avatar, ruleset cover badge, editor cover pane). Measured across all six band stories: the portrait card face renders 96x135 uncropped while the four squares hold 96x96.

asset card, full face, fixed

For the record on why my proof missed it: the twenty-shot matrix used `pages-assets--deck`, whose seeded deck face happens to render near-square, so the crop had nothing tall to bite — the treachery-card story above is the state where the change could plausibly break, and it is now the story this claim rides on. The proof-selection lesson (shoot the states that can break, not the states you already shot) is already in my memory from #716; this is it biting a second way: vary the SUBJECT's shape, not only the viewport. CI re-running; audit stays red until the #799 operator steps land, everything else expected green.

Issue #814, opened 2026-08-27
The shared ingest secret retires


ndelangen:
## Question

Once covers and avatars both ride the ledger path, the shared secret has no callers:

- Remove the bearer-secret support from the Worker's ingest endpoint and from the Convex side.
- Delete USER_IMAGE_INGEST_SECRET from the Worker's secrets and the Convex env (both deployments); these are Norbert-or-CI's hands, never a session's.
- Update the audit gate's Cloudflare live-drift contract in the same change, so the gate expects the secret's absence rather than flagging it.

## Coordination

Waits on the ledger build and on avatars riding it; the secret retires last.

ndelangen:
Sequencing constraint from #813's delivery: the cover backfill (rulesetCovers:backfillLegacyCovers) still rides the bearer path and has not yet run in prod. Either that run happens before this ticket executes, or this ticket's scope grows to migrate the cover backfill onto the token path first. The avatar backfill already rides tokens and does not constrain this.

ndelangen:
Delivered whole. The code half merged as PR #824 and deployed; the operator half ran on 2026-08-28: USER_IMAGE_INGEST_SECRET is deleted from the live Worker (verified by a names-only secret list showing two entries) and removed from both Convex deployments per Norbert. The audit gate re-ran green against the two-secret contract, which is the end-to-end proof: no code references the secret, no store holds it, and the gate that watches the live account agrees. The ingest's only credential is now the per-request ledger token.

Issue #813, opened 2026-08-27
Avatars rehost on save through the ledger path


ndelangen:
## Question

Implement rehost-on-save for profile avatars, mirroring the cover flow, on the ledger-token path:

- Profile save with a changed `avatar_url` mints a ledger token and calls the Worker; the Worker fetches, encodes the decided renditions, stores content-addressed in R2, and calls back with the result plus the token.
- The seeding path from the auth provider's image is covered or explicitly deferred per the renditions ticket's decision.
- The prod backfill of existing avatar rows is the closing operator step (Norbert's hands), mirroring the cover backfill with its rehosted, skipped, failed, truncated summary.

## Coordination

Waits on [Ingest auth graduates to Convex-ledger tokens](https://github.com/ndelangen/dunezone/issues/809) for the ledger path and on the avatar-renditions decision for the shapes.

ndelangen:
Delivered end-to-end. Built as PR #818 (merged as cb63491c17a, orchestrator-reviewed at c47f248d7dd), deploy-main succeeded on the merge commit: the avatar pipeline is live with its 320px square rendition, the ledger's profile_avatar capability, the superseded race guard, and the async mint-and-fire on every external write path. The one remaining operator step is the prod backfill, npx convex run profileAvatars:backfillLegacyAvatars --prod, which reports {rehosted, skipped, failed, truncated}. The pending-window UX question lives in [Does the profile editor surface a pending rehost](https://github.com/ndelangen/dunezone/issues/816); the secret's retirement in [The shared ingest secret retires](https://github.com/ndelangen/dunezone/issues/814) is now unblocked, with one sequencing constraint recorded there.

Issue #812, opened 2026-08-27
Decide the avatar renditions and the shape that carries them


ndelangen:
## Question

Covers settled on two renditions (full 1600px progressive, thumb 320px) carried as a `cover` object with the legacy `image_cover` echo. Avatars need their own answers before the build:

1. **Rendition set.** Which sizes does avatar use warrant (a 320px thumb alone? a smaller chip-size too?), and is the square crop enforced at encode time or left to CSS as today?
2. **Schema shape.** Does the profile row grow an `avatar` object mirroring `cover` (`{url, thumb_url}` and friends) with `avatar_url` kept as the legacy echo, or something smaller?
3. **The seeding path.** Profile creation copies the auth provider's image into `avatar_url` (convex/profiles.ts). Does that seed get rehosted at creation time, lazily on first save, or stay external until the user touches the field?

ndelangen:
Resolved with Norbert, 2026-08-28. Three decisions:

1. **One 320px square rendition**, center-cropped at encode. The largest avatar render in the app is the profile band at 96px CSS (192px at 2x DPR), so 320 covers every site with headroom; squareness becomes a property of the stored asset and consumers stop handling non-square sources.
2. **Schema: the cover shape minus thumb.** `avatar: {url, source_url, width, height}`, optional and nullable while pre-rehost rows exist, written only by the pipeline; `avatar_url` stays as the dual-written legacy echo until a later retirement, mirroring `image_cover`. Same invariants and backfill pattern as covers.
3. **Rehost every external write.** The creation seed from the auth provider, the edit save, and the prod backfill all funnel through the same mint-and-fire; no external URL persists past its callback window, and the profile renders the external URL only until the callback lands.

A consequence for [Avatars rehost on save through the ledger path](https://github.com/ndelangen/dunezone/issues/813): with a single rendition the display side reduces to one fallback accessor (`avatar.url ?? avatar_url`) in `src/app/db/profiles.ts`, so the display-sweep fog collapses into that ticket rather than becoming its own.

Issue #811, opened 2026-08-27
Research the Convex primitives for the token ledger


ndelangen:
## Question

Primary-source research (Convex docs and reference) that the ledger build in #809 waits on:

1. **Exposure of the introspection read.** The Worker asks Convex "is this token valid?" before doing work. What is the right Convex primitive for an unauthenticated external read: an `httpAction`, or a public query via the HTTP client? What rate or abuse behaviour does each have, and does Convex offer anything (built-in limits, recommended patterns) for a public endpoint whose job is answering strangers cheaply?
2. **Ledger TTL hygiene.** Rows are `{id, capability, expires, consumed}`. What is the recommended pattern for expiring and deleting stale rows: scheduled functions, crons, lazy deletion on read? What guarantees do Convex scheduled functions give here?
3. **Callback payload shape.** The Worker's single callback delivers the result and consumes the token in one mutation. Should that payload also carry the R2 object keys, so [User-image ingest hardening](https://github.com/ndelangen/dunezone/issues/801)'s future GC can know what an ingest produced? Weigh against keeping the callback minimal.

Findings land as a comment on this ticket with citations to the primary sources; no repo commits.

ndelangen:
Research findings from the Convex docs and the get-convex GitHub org. Sources are cited per section; no repo changes made.

## 1. Exposure of the introspection read

**Direct answer:** use a public query called through Convex's built-in HTTP API for the "is this token valid?" read, and a public mutation through the same API for the validate-and-write callback. An httpAction wrapper adds cost and no security here. The atomicity you need for read-check-then-consume in one mutation is explicitly guaranteed.

How each is exposed. Every public Convex function is already an HTTP endpoint: POST to `{CONVEX_URL}/api/query`, `/api/mutation`, or `/api/run/{functionPath}` with a JSON body of `{path, args}`. Public functions need no authentication over this API (https://docs.convex.dev/http-api/). httpActions instead live on a separate domain, `https://<deployment>.convex.site`, with routes you define yourself, and you handle CORS manually (https://docs.convex.dev/functions/http-actions). CORS is irrelevant for us either way since the caller is a Worker, not a browser.

[...]

PR #815, opened 2026-08-27
The cover ingest authenticates with single-use ledger tokens the Worker introspects

ndelangen:
Builds #809. Design per the in-issue decision, the research findings on #811, and the decisions comment on #812.

## What changed

The user-image ingest stops trusting a long-lived shared secret on the author path and moves to token introspection with a Convex ledger. The Worker holds nothing; possession of a single-use, unguessable token is the whole credential, and the ledger is asked before any expensive work happens.

**Convex**

- New `user_image_ingest_tokens` table: `token_id` (256 random bits as hex from the runtime crypto source), `capability` (which entity and field this ingest may set, a single-member union today: `ruleset_cover` plus the ruleset id), `source_url` (pinned at mint so a token holder cannot rewrite provenance), `expires` (ms, authoritative on every read), `consumed`, and `r2_keys` recorded at consume time. Indexed for point lookup by `token_id`.
- New `convex/ingestTokens.ts` module:
  - `mint` (internal): inserts the row and schedules its own deletion at `expires` via `ctx.scheduler.runAt`, atomic with the mint and exactly-once, so no cron sweeps the table.
  - `check` (public query, called by the Worker over `/api/query`): indexed point read answering valid or invalid. Never consumes, reveals only the boolean. It takes `now` as an argument because the repo's Convex guidelines forbid wall-clock reads in queries; a hostile caller lying about the time only lies to itself, since this answer gates nothing on the server. The consume mutation re-checks expiry against the server clock and is the only gate that counts.
  - `consume` (public mutation, called over `/api/mutation`): in one transaction it re-reads the token row, rejects missing, expired, or consumed tokens, validates the payload against a strict shared Zod floor (delivery URLs restricted to our namespace, dimensions inside the encoder box, at most four content-addressed keys), writes the cover where the capability points, records the R2 keys, and marks the row consumed. Double consume is prevented by Convex serializability alone; there are no locks.
- `rulesetCovers.rehost` now runs validate, access check, config check, mint, then the Worker POST with the token and no bearer header. The response is a completion signal only (`{ok: true}`); the write arrives through the consuming mutation. The save-and-wait UX is unchanged. The old `commit` internal mutation is gone; `patchStoredCover` moved to `convex/lib/rulesetCover.ts` and is shared by the backfill commit and the ledger consume.
- The operator backfill (`backfillLegacyCovers`) stays on the bearer-secret path unchanged; that path retires with the secret in #814.

**Worker** (`workers/publisher/user-images.ts`)

- Dispatch: a request with a bearer header takes the legacy path unchanged; anything else must carry a token in the body. A tokenless, bearerless request is a 401 after one bounded body read.
- Token path: ask `ingestTokens:check` first (invalid answers 403 before any fetch), then the existing fetch, encode, and store pipeline with every safety property intact (https-only source, hand-walked redirects, 10s timeout, 10MB streaming cap, content-addressed keys), then `ingestTokens:consume` with the result and both R2 keys. A bounced consume surfaces as a 409 with an author-facing reason; an unreachable ledger is 503 before spending anything.
- The Convex deployment origin arrives as a new `CONVEX_CLOUD_BASE_URL` wrangler var, configuration rather than request data, so a caller cannot point the Worker at a ledger that always says yes. The reviewed deploy contract and the live-drift fixture were extended for the new binding.

## The flow, in prose

Author saves a cover. The action validates the URL, checks edit access, mints a ledger row with a scheduled self-deletion, and POSTs the source URL plus the token to the Worker without any secret. The Worker asks the ledger whether the token is live; only a yes unlocks the fetch, encode, and store work. The Worker then calls the consuming mutation, which in one transaction validates the payload, writes the cover and its legacy dual-write to the ruleset the capability names, records the two R2 keys, and burns the token. The Worker answers the still-waiting action with a bare completion signal, and the author sees success or an honest refusal. A replayed callback meets the consumed tombstone and bounces; a failed fetch never burns the token; an expired unconsumed row deletes itself.

## Calls made in this PR (flagging for review)

- **R2 key record**: consumed rows are kept as tombstones carrying `r2_keys`, so future GC (#801) walks the ledger with an indexed read. Only unconsumed rows are deleted at expiry. No separate ledger table.
- **`check` takes `now` as an argument** (see above); this is the one place the shape deviates from the issue text, forced by the no-wall-clock-in-queries guideline.
- **`convex/_generated/api.d.ts` was extended by hand** with the two new module entries in the file's own deterministic pattern. Running codegen was off the table: the shared-dev rule forbids it here, the worktree has no `.env.local`, and the running Docker backend belongs to another session. `dataModel.d.ts` is schema-generic and needed no change; `server.d.ts`/`server.js` (the pinned `env` export from #807) are untouched.
- **A consume whose target row is gone** still records the keys on the tombstone and reports `entity_gone`, so the bucket objects stay on the GC record.
- **Token TTL is 15 minutes**, generous headroom over a job that finishes in seconds.

## Out of scope, per the map

Avatars (#813), secret removal (#814), fire-and-forget UX (stays in fog per the #809 comment), and the #801 hardening items.

## Gates

| Gate | Result |
| --- | --- |
| `bun run lint` | exit 0 |
| `bun run format:check` | exit 0 |
| `bun run typecheck` | exit 0 |
| `bun run publisher:typecheck` | exit 0 |
| `bun run test` | exit 0, 811 passed (7 new ledger tests, 6 new Worker token-path tests, 2 cover-flow tests rewritten for the callback write) |
| `bun run publisher:release:verify` | exit 0, no manifest diff |
| `check:prose`, `check:app-layout`, `check:css-orphans`, `check:convex-skip` | exit 0 |

New coverage: valid token unlocks, expired rejects (check and consume), consumed rejects, double-consume race with one winner, check never consumes, payload floor holds without burning the token, the cover flow writes through the callback, and the response body alone writes nothing.


ndelangen:
Orchestrator review, anchored to head `029777f4e3c` (the worker's build `2d264043bc8` plus one disclosed review commit of mine, see below).

**Verified against the spec.** The build matches #809's decided design, #811's research and #812's shapes at every point I checked: the check is a public query and the callback a public mutation over the built-in HTTP API; mint schedules its own row deletion, exactly-once and atomic with the insert; consume validates, writes where the capability points, records the R2 keys and burns the token in one transaction, relying on serializability alone for the double-consume race; consumed rows tombstone with their keys as the GC record. `source_url` is pinned at mint, so a token holder cannot rewrite provenance. The Worker refuses a dead token (403) or an unanswered ledger (503) before spending a fetch, and the ledger origin is configuration, never request data.

**Gates re-verified independently.** I re-ran lint, format:check, typecheck, publisher:typecheck and the test suite in the build worktree at both the original and current head, each read from a bare exit code: all 0, 811 tests passing, 122 files. The "response body alone writes nothing" test discriminates properly: a Worker answering the retired result shape both throws and leaves the row bare.

**The `now` argument on `check`** is compliance, not a workaround: the repo's Convex guidelines forbid wall-clock reads in queries and prescribe exactly this pattern. `consume` re-checks expiry on the server clock and is the only gate that counts.

**One property change, accepted.** The old `commit` re-checked edit access at write time; now access is checked once at mint and the capability authorizes the write for the token's fifteen-minute life. That is inherent to the ledger design (the Worker has no viewer identity) and the JSDoc states it.

**My review commit `029777f4e3c`,** disclosed for provenance since all sessions share one git identity: the delivery-URL floor's comment claimed to stop arbitrary foreign URLs while pinning only protocol and path shape, and now says what it actually pins (host pinning is recorded on #801); and the test fetch stub's prefix matching, which CodeQL flagged twice as incomplete URL sanitization, now compares the parsed origin exactly. Both CodeQL alerts should clear on this head.

**Residual risks, named.**
1. `convex/_generated/api.d.ts` was extended by hand because codegen was rightly off-limits in the worktree. The edit follows the file's own deterministic pattern exactly (I checked ordering and quoting against every neighbour), but byte-parity with real codegen output is unproven until deploy-main's exact-checkout preflight runs post-merge. The failure mode is a refused deploy, not a bad one, and the fix pattern is known from #807.
2. Between merge and the deploy-main run, the audit gate's live-drift check will see the live Worker missing `CONVEX_CLOUD_BASE_URL` and go red. It is a wrangler var, not a secret, so the deploy itself provisions it and the gate self-heals. No operator step needed.

TTL at fifteen minutes, the 409/503 status mapping and bearer-first dispatch when both credentials appear were all reviewed and are fine as built. Review stands for this head only; it lapses if the branch moves.

Issue #803, opened 2026-08-27
Two login-gate spellings differ during the pending window


ndelangen:
## Question

Found during the asset-editor conversion (#802): the app gates signed-out readers with two spellings that behave differently while the profile query is pending. `profile.data === null` treats pending as signed-in and shows the editor briefly to an anonymous visitor before the gate lands; `!profile.data?.x` treats pending as gated and shows the gate immediately. Both resolve correctly once the query settles.

Decide the one spelling (the eager gate is the likely answer: a flash of editor at a signed-out reader is worse than a flash of gate at a signed-in one), sweep the gates onto it, and consider whether the session contract should make the wrong spelling unwritable rather than merely absent.

ndelangen:
Counted during the composition wave's audit, against main at `ac8a223b2be`. There are three spellings, not two, and the third is the one that varies most.

**Eager, distinguishes unresolved from signed out.** `profile.data === null`, five sites, all asset create pages: `-treacheryCreate.tsx:56`, `-bundleCreate.tsx:57`, `-tokenCreate.tsx:62`, `-deckCreate.tsx:64`, `-rectangleCreate.tsx:64`.

**Truthy, gates during the pending window too.** `!profile.data`, one site: `profiles/$profileSlug/edit.tsx:448`.

**Field-probing, and each site picks a different field.** `admin/migrations.tsx:38` reads `_id`, `rulesets/create.tsx:101` reads `user_id`, `groups/create.tsx:30` reads `_id` and `slug` together. Same intent, three different fields standing in for "is there a profile".

Unaffected by all of this: the routes that gate on server-derived `viewerAccess.viewer.kind`, which is why `Assets/EditDeckSignedOut` is coverable and the nine above are not.

One consequence worth weighing when this is settled, from [#769](https://github.com/ndelangen/dunezone/issues/769): the page-story seam's `parameters.identity: null` leaves `profiles.session` unresolved rather than answering signed out, so a `=== null` gate renders its editor in a story while gating correctly in production. Picking the eager spelling makes that gap matter for the whole list; picking a spelling that treats unresolved as not-signed-in would close it. The story coverage and the spelling decision are the same decision.

ndelangen:
Resolved with Norbert, 2026-08-28: **tri-state via one derived helper.**

[...]

PR #817, opened 2026-08-27
Four page stories assert behaviour the page itself owns

### LINKED ISSUE #769: Add selective behavioural validation to page stories

ndelangen:
Closes [#769](https://github.com/ndelangen/dunezone/issues/769).

Four page stories gain assertions. No source file changes.

## How the four were chosen

Every page story was surveyed against the ticket's rules, and every candidate was then handed to an independent reviewer told to refute it: does the behaviour exist, is it page-owned rather than a widget's, would the assertion pass even with the behaviour removed, is the tier right, and would this read as contract-level coverage or as ceremony. Thirty-three candidates were judged. Sixteen survived, seventeen were refuted, and four were built.

The seventeen refusals did more for this PR than the sixteen survivors. Several died on the ceremony bar rather than on mechanics, which is the ticket's actual constraint, and one died because the repo had already ruled against that exact play in writing.

## What was built

| story | behaviour kind | the assertion, and why it is not the obvious one |
|---|---|---|
| `Assets/CreateTreacheryCardSaves` | primary mutation, then navigation | Asserts a `Delete card` control, which only the editor carries. The toolbar badge would read "Saved" from a resolved promise alone; the route cannot reach the edit page without a real id and slug coming back. |
| `Assets/CreateTreacheryCardChapterJump` | mode switching | Asserts the Body field is absent, then present after the chip. The chapter routing is the route's own: `ValidationHeader` only reports which warning was chosen and the page decides `setChapter(warning.chapter)`. |
| `Assets/EditDeckMemberCountWrites` | mutation-driven query refresh | Asserts the rail total, **not** the count field. `MemberCountInput` keeps what you typed in its own state and reclaims it on commit, so asserting the field would pass whether or not anything was written. The total is reduced from the members the page query returns. |
| `Rulesets/AskQuestionMissingRuleset` | the route's own error frame | Gains the assertions it shipped without in #800, where I verified it by hand instead. Asserts the alert's own title and that the way back lives outside `main`, because the router's default renders too. |

## Every one was run against the behaviour removed

Not reasoned about. Each behaviour was deliberately broken in the source, the suite re-run, and the source restored.

| story | with the behaviour | with it removed |
|---|---|---|
| `CreateTreacheryCardSaves` | pass | **fail** (navigate disabled in `onSuccess`) |
| `CreateTreacheryCardChapterJump` | pass | **fail** (`onFocusWarning` made a no-op) |
| `EditDeckMemberCountWrites` | pass | **fail** (`onCountChange` mutation disabled) |
| `AskQuestionMissingRuleset` | pass | **fail** (`errorComponent` removed) |

## Deliberate omissions

The two the ticket and the review specifically asked about:

**The homepage's staged pending and error stories, deferred here from #797, are not built.** They are feasible without seam work, so they do not trip the split condition, but they are not worth building: `PageMessage.stories.tsx` already exports `Loading`, `Failed` and `NoWayBack`, which are the exact frames the homepage uses, and the page-owned residue is two lines of route config shared by eleven routes for `errorComponent` and five for `pendingComponent`. Verified both counts. If that wiring is worth guarding it wants one cheap check across all sixteen, not two mock-driven stories on the one route where nothing page-owned happens. Raising that shape rather than building these.

**Login gates that read the profile session remain uncoverable.** `parameters.identity: null` leaves `profiles.session` unresolved rather than answering signed out, so a route gating on `profile.data === null` renders its editor in a story while gating correctly in production. The affected pages are the five asset create routes, `rulesets/create`, `groups/create`, `admin/migrations` and `profiles/$profileSlug/edit`. Routes gating on `!profile.data?.x` or on server-derived `viewerAccess.viewer.kind` are unaffected, which is why `Assets/EditDeckSignedOut` works. This list is also posted on #769 so it survives this PR's merge as tracked work.

Others, with the reason rather than a shrug:

- **Held deletion** is not played. The repo already decided this: `ConfirmDeleteAction.stories.tsx` says the full hold is pinned by the unit suite with fake timers because a story should not sit through it.
- **Group membership moderation** would need story-local seeding of a second user and a pending row, and what it would prove is a mutation-driven refresh already covered on the deck editor.
- **`/__icons` search** is unlinked developer tooling, reads nothing from the seeded database, and is the archetypal read-only story the ticket says to leave alone.
- **The other four asset editors and four create twins** are omitted on purpose: they differ from the covered ones in their widget and their nouns, not in their route-owned behaviour. Covering the contract once is the contract; covering it five times is the matrix the ticket rules out.
- **Rename-and-follow-slug** appears on four pages. It is not covered because the observable in a story is the page's own name refreshing from the query, which is the kind already covered, and the slug change itself is not observable without asserting the URL.

## Verified

Gates run individually and read from bare exit codes, not through a pipe: `lint` 0, `format:check` 0, `typecheck` 0, `publisher:typecheck` 0, `test` 0 (767), `storybook:test` 0 (460, up 4), `check:prose` 0, `knip` 0, `check:app-layout` 0, `check:css-orphans` 0, `check` 0 with no capture-bundle changes.

Worktree off `e50938b2df4` with its own install. The end-to-end suite is untouched.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for creating and saving cards, navigating to relevant chapters from validation warnings, and updating deck member counts.
  * Added coverage for missing ruleset errors, including visible messaging and back-link placement.

### DIFF
--- a/src/app/routes/-assets.stories.tsx
+++ b/src/app/routes/-assets.stories.tsx
@@ -1,5 +1,5 @@
 import preview from '@sb/preview';
-import { expect, within } from 'storybook/test';
+import { expect, userEvent, within } from 'storybook/test';
 
 import { pageStoryMeta } from './-storybookConfig';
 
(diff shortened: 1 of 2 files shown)


ndelangen:
Orchestrator review, anchored to head `396dfac5157`, the branch's single commit. I read the whole diff rather than the summary of it.

**The four stories check out.** Each assertion is discriminating by structure, not just by the A/B table's word: the save story's proof is a control only the destination page carries; the chapter jump asserts absence before presence; the deck story asserts a total reduced from the page query rather than the field that would lie; and the ask-a-question frame asserts the alert's own title plus a way back living outside `main`, which the router's default cannot produce. The A/B table names a specific sabotage per story, which is the practice we want on the record. Comment style and story conventions match the house rules.

**The homepage omission is accepted, and the ruling goes one step further.** The deferred staged stories would re-prove `PageMessage`'s own exported frames through a route that adds nothing page-owned; that is the exact duplicate-proof ceremony the ticket rules out, and the deviation from my earlier "ride along" ruling is correct on the merits. The proposed alternative, one config-presence check across the sixteen routes, is also declined: `router.tsx` sets a `defaultErrorComponent`, so a route without explicit wiring falls back by design rather than standing unguarded, and a check requiring the explicit lines would encode a non-invariant. The wiring stays a review-time read. I will note the disposition on #797 so the deferral does not dangle.

**Scope-down is honest and verified.** The uncoverable-pages list is on #769 with the mechanism named, tied to #803 where settling the gate spelling would dissolve most of it. The seventeen refuted candidates with reasons are worth more than seventeen built stories would have been.

CI is the independent gate run for this stories-only diff; the storybook job and CodeRabbit's review are still pending as I write this, so the merge call waits for them. Review stands for this head only and lapses if the branch moves.

Issue #816, opened 2026-08-27
Does the profile editor surface a pending rehost


ndelangen:
## Question

Covers keep save-and-wait, so the author sees success or failure before the dialog closes. Avatars are async by decision: the mutation writes the external URL, the rehost lands later through the callback, and the profile renders the external URL until then.

Does the profile editor surface that pending window (a subtle "storing your image" state, resolved on the callback), or is the silent flip acceptable? If pending UI is wanted, what does it look like and where does it live? The avatar build ships with the silent flip; this decision, if it wants more, becomes a small follow-up on the editor.

ndelangen:
Resolved with Norbert, 2026-08-28: the silent flip stands, with no pending UI and no build. The grounds: the window lasts seconds (the Worker's whole budget is a 10s source fetch plus encode and store), the pasted external URL renders throughout it so nothing is ever visibly broken, and a failed rehost degrades to exactly the pre-pipeline behaviour, an external hotlink that keeps rendering, logged but not shown. Revisit only if a real support burden appears; the failure-feedback option in this ticket's discussion is the shape to reach for then.

PR #818, opened 2026-08-28
Profile avatars rehost on save through the ledger token path

ndelangen:
## What changed and why

Profile avatars now rehost through the same Convex-ledger token path the cover pipeline established in #815, per the decisions on #812: one 320px square rendition, center-cropped at encode time, stored content-addressed, with `avatar: {url, source_url, width, height}` on the profile row and `avatar_url` kept as the dual-written legacy echo.

## The flow in prose

A profile save with a genuinely new external URL writes that URL as `avatar_url`, clears any stored `avatar`, and schedules `profileAvatars:rehost` in the same transaction. The scheduled action validates the URL shape, mints a single-use token with the new `profile_avatar` capability, and POSTs to the Worker with no author waiting. The Worker asks the ledger's `check`, which now answers `{valid, kind}`; a `profile_avatar` kind selects the square center-crop recipe (`fit: 'cover'` into a 320px box, progressive JPEG, minimum-edge and squareness refusals). The stored result lands through the ledger's `consume`, whose avatar arm holds its own Zod floor, rechecks that the row's `avatar_url` still equals the token's pinned source, and writes the avatar plus its echo via `patchStoredAvatar`. Until that callback flips it, every reader renders the external URL; no pending-state UI exists (#816 owns anything more).

The creation seed funnels through the same path: `ensureProfileForUser` schedules a rehost for any profile it leaves holding an external URL with no stored avatar, which covers the OAuth callback seed, `bootstrapCurrent`, and legacy rows touched by a save. The operator backfill (`profileAvatars:backfillLegacyAvatars`) scans the oldest 500 profiles for external URLs without a stored avatar, skips deleted-state accounts, and mints and POSTs per row on the token path, reporting `{rehosted, skipped, failed, truncated}`.

## Resolved design points

- **Recipe from the ledger.** Took the recommended mechanism: `check` returns `{valid, kind}`, with `kind` present only on a live token. The Worker treats a live answer without a kind as the cover recipe, which keeps the deploy window safe (a ledger predating kinds only ever minted cover tokens); the existing token-path Worker test proves that default. The bearer path stays cover-only.
- **Capability-aware consume without weakening covers.** The Convex-level arg validator makes `result.thumb_url` optional; each capability arm then parses its own strict Zod floor. The cover arm is byte-for-byte the old schema. The avatar arm (`userImageAvatarIngestCallbackSchema`) demands one delivery URL, integer dims bounded by 320 and equal to each other, and exactly one content-addressed key; it is a `strictObject`, so a payload carrying the cover shape's `thumb_url` is refused rather than trimmed. The token row is now read before the payload parse (the row's capability decides which floor applies); a refused payload still does not burn the token.
- **The race guard lives at the write site.** Because the backfill rides the token path, the write happens inside `consume`, so the expected-echo recheck moved there: a callback whose pinned source no longer matches the row's `avatar_url` burns the token into its tombstone (the R2 keys stay on the GC record) and returns a new `superseded` refusal without touching the row. This one guard protects both the backfill race and rapid consecutive saves on the async path. The backfill classifies a failed row as `skipped` when the row's echo moved while the Worker was fetching, `failed` otherwise.
- **Where scheduling lives.** The issue names `createProfileIfMissing`, but the actual OAuth seed runs through `auth.ts`'s `afterUserCreatedOrUpdated`, which calls `ensureProfileForUser` directly. The scheduling hook (`scheduleAvatarRehostIfPending`) therefore sits inside `ensureProfileForUser`, one level down, where every seed funnels; `updateCurrent` calls it additionally after writing a new external URL.
- **Echo guard semantics.** The edit form round-trips `avatar_url` (after a rehost, our delivery URL). A submission equal to the stored avatar's `url` or to the current `avatar_url` neither mints nor touches `avatar`; empty and null never mint (the form schema already floors the field non-empty).
- **Shared plumbing.** `ingestBaseUrl`, `throwIngestFailure` and `ingestWithToken` moved from `rulesetCovers.ts` to `convex/lib/userImageIngest.ts` since both pipelines use them; their messages were neutralized from "Cover storage ..." to "Image storage ...". The bearer-path helpers stay in `rulesetCovers.ts` and retire with the secret.

## Display side

One accessor, `profileAvatarUrl` in `src/app/db/profiles.ts` (`avatar.url ?? avatar_url`), mirroring `coverUrl`. Full-row consumers route through it: the SiteNavigation account menu, the profiles list, the profile detail band, and the faction detail owner chip. Flat server projections (profile summary chips, roster entries, discovery, the replacement-profile picker) apply the same fallback at projection time, so `ProfileLink`, `FaqList`, `Spotlight` callers and `ProfilePicker` need no membrane change. Page contracts extend automatically: the profile validators derive from the schema (`docValidator('profiles')`), so `avatar` rides every profile-shaped return.

**Visual parity:** rendered output is unchanged by construction: no fixture or existing row carries an `avatar` object, so the accessor returns exactly the old `avatar_url` for all current data, and after a rehost the dual-written echo keeps both expressions equal. Verified three ways: the Storybook suite (`bun run storybook:test`, 457 stories, all passing), and a live Storybook pass over `Content/ProfileLink` (WithAvatar renders `` exactly as its args say; InitialsFallback still renders "CE" + name) and `Shell/SiteNavigation` (signed-out stories render the Login slot; the signed-in avatar branch computes the identical value for any avatar-less row).

## Tests

- `convex/ingestTokens.ledger.test.ts`: avatar `check` kind, the avatar consume floor (rectangular, oversized, two keys, cover-shaped payloads all refused without burning; a good consume still lands), and the `superseded` bounce with its tombstone.
- `convex/profiles.avatar.test.ts` (new): the async save flow (external URL renders at once, scheduled callback flips it, token in the body, no bearer header), the echo guard (no mint, no fetch, avatar untouched), the creation seed (identity `pictureUrl` lands without a save), the backfill summary (rehosted/failed, deleted and settled rows unscanned), and the backfill race (user's mid-fetch change wins, counted skipped).
- `workers/publisher/user-images.test.ts`: an avatar token selects `{width: 320, height: 320, fit: 'cover'}`, stores one object, consumes one key with no thumb fields; a non-square encode is refused before storing or consuming.

Every new error-state assertion was A/B verified: the square Zod refine, the consume `superseded` guard, the mutation echo guard, the seed scheduling, the backfill skipped-classification and the Worker's square refusal were each removed in turn and their test watched to fail before being trusted.

## House-rule disclosures

- No convex CLI ran. `convex/_generated/api.d.ts` was hand-extended for the three new modules (`lib/profileAvatar`, `lib/userImageIngest`, `profileAvatars`) in the generator's deterministic pattern: case-sensitive ASCII placement in both the import block and the `fullApi` keys, double-quoted `.js` paths, quoted keys only for `lib/` entries. `server.d.ts`/`server.js` untouched. Same procedure PR #815 used and deploy-main's preflight proved.
- `bun run publisher:release:verify` ran with `VITE_CONVEX_URL` set to the checked-in production constant from `scripts/publisher-deployment-contract.ts`; exit 0, no `renderer-manifest.generated.ts` diff.
- Schema change is additive (`avatar` optional); no migration needed. Norbert runs `profileAvatars:backfillLegacyAvatars` in prod as the closing operator step.

## Gates (bare exit codes)

- `bun run lint`: 0
- `bun run format:check`: 0
- `bun run typecheck`: 0
- `bun run publisher:typecheck`: 0
- `bun run test`: 0 (123 files, 821 tests)
- `bun run storybook:test`: 0 (116 files, 457 tests)
- `bun run publisher:release:verify`: 0

Builds #813


ndelangen:
Orchestrator review, anchored to head `c47f248d7dd`: the worker's two build commits plus one disclosed review commit of mine. I read the full diff and the interaction seams rather than the report of them.

**Faithful to the decided shape.** One 320px square center-cropped at encode with the encoder's `cover` fit and a non-square result refused twice (Worker refusal with a readable message, consume floor as the backstop); `avatar: {url, source_url, width, height}` written only by the ledger consume with `avatar_url` dual-written as the legacy echo; and every external write funnels through mint-and-fire: the edit save, the sign-in seed, and the backfill.

**The trust boundary holds.** Consume's arms each parse their own strict floor before any write, a cover-shaped payload on an avatar token is refused whole, and a refused payload never burns the token. The new `superseded` recheck lands after the tombstone patch, so a stale callback still leaves its R2 keys on the GC record while the newer write keeps the row. The check query's `kind` rides only on a live answer and names a recipe, not an entity.

**The two judgment calls I verified rather than accepted:**
- *The seed moved to `ensureProfileForUser`*, deviating from the ticket's named hook. Correct: that is the funnel every seed actually passes through, and `refreshExistingProfile` only fills a null `avatar_url`, never overwrites one, and leaves a cleared empty string cleared, so the schedule guard cannot fight a custom avatar or resurrect a removed one.
- *The echo guard matrix in `updateCurrent`*: a round-tripped delivery URL does not mint, a double-submit of the same external URL mid-flight does not double-mint, a genuinely new URL nulls the stored avatar so the page honestly renders the submitted URL until the callback flips it, and clearing the field schedules nothing.

**My review commit `c47f248d7dd`, disclosed:** CI's knip gate caught `USER_IMAGE_INGEST_KINDS` exported but consumed only inside the contract file; it is now module-local. The gate-parity miss was mine, not the worker's: knip was absent from the build brief's gate list, and it is in CI's. Future briefs carry it.

**Gates re-verified independently** in the build worktree at this head, each from a bare exit code: knip, lint, format:check, typecheck, publisher:typecheck and the suite all 0, 821 tests across 123 files.

**Discussion points, adjudicated:**
- A legacy row whose external URL is permanently dead re-schedules one rehost per sign-in, forever. Accepted: one bounded fetch per sign-in event, convergence on first success, and the suppression state a retry-limit would need is not worth carrying for the handful of rows the backfill will not already convert.
- The edit form showing the pasted URL until remount while the row's echo has flipped: accepted as state-beside-a-draft territory; the next untouched save costs one convergent re-mint and nothing else.
- CodeScene's two advisory flags stand as-is: a 0.02 hotspot decline on the faction page for routing its owner chip through the accessor, and the backfill loop's shape, which mirrors the cover backfill's decided form.

Awaiting CI on this head before the merge call. Review stands for `c47f248d7dd` only and lapses if the branch moves.

